PR #136, opened 2022-04-20
runner: remove the trailing endline from stdout printing

gboutry:
By default, python prints a newline at the end of the `print` function. Remove the endline because ansible already puts endlines itself.

This PR improves stdout readability.


rpignolet:
Oh yeah well done to have found the problem with empty newline in Ansible log.

rpignolet (approved):

PR #147, opened 2022-04-28
core: timestamp without timezone, datetime from utc

### LINKED ISSUE #81: database: decide how to manage timestamps
Currently, we use simple timestamps without timezones in postgres and datetime in sqlite. And the timestamp inserted in the database is a simple python `datetime.now()` without any after thought.

This could become a source of problem in different areas of the world, such as summer time vs winter time.

gboutry:
fix #81 

In this PR:
- explicitly set Timestamp without Timezone
- use `utcnow()`, instead of `now()`

### DIFF
--- a/tdp/core/models/action_log.py
+++ b/tdp/core/models/action_log.py
@@ -14,8 +14,8 @@ class ActionLog(Base):
 
     deployment_id = Column(Integer, ForeignKey("deployment_log.id"), primary_key=True)
     action = Column(String(length=NODE_NAME_MAX_LENGTH), primary_key=True)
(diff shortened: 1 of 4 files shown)


rpignolet:
You do not modify the `browse` command to display time with local timezone ?

gboutry:
I think it should be part of another PR.

I'm modifying the core of the library.

Then, I'll change the clients.

rpignolet (approved):

gboutry:
After some thinking, I think you're right, it's better to correct this in the same PR

rpignolet (approved):

PR #148, opened 2022-04-28
core: prevent empty commits in git_repository

### LINKED ISSUE #143: Configuration changes without actual changes makes an empty commit
We need to prevent the lib from making an empty commit when no change is actually performed.

leopaul36:
Fix #143 

I am really not sure this is the best way to do this thought.

**Note: ** I had to use `--no-verify` to commit as `isort` and `black` seem to counter each other. Runnin the commands gives the same output over and over:

```
isort --recursive tdp && black tdp
Fixing /home/leo/TOSIT-FR/tdp-lib/tdp/core/repository/test_git_repository.py
Fixing /home/leo/TOSIT-FR/tdp-lib/tdp/core/repository/git_repository.py
reformatted tdp/core/repository/git_repository.py
reformatted tdp/core/repository/test_git_repository.py
All done! ✨ 🍰 ✨
2 files reformatted, 35 files left unchanged.
```

### DIFF
--- a/tdp/core/repository/git_repository.py
+++ b/tdp/core/repository/git_repository.py
@@ -4,9 +4,14 @@
 import logging
 from contextlib import contextmanager
 
-from git import InvalidGitRepositoryError, NoSuchPathError, Repo
+from git import BadName, InvalidGitRepositoryError, NoSuchPathError, Repo
 
-from tdp.core.repository.repository import NotARepository, NoVersionYet, Repository
+from tdp.core.repository.repository import (
+    EmptyCommit,
+    NotARepository,
+    NoVersionYet,
+    Repository,
+)
 
 logger = logging.getLogger("tdp").getChild("git_repository")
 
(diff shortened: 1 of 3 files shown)


gboutry (changes requested):

tdp/core/repository/git_repository.py
```diff
@@ -41,6 +46,13 @@ def init(path):
...
             yield self
+            try:
+                if len(self._repo.index.diff("HEAD")) == 0:
+                    raise EmptyCommit("The commit has no diff")
+            except BadName as e:
+                logger.debug(
+                    f"error during diff: {e}. Probably because the repo is still empty."
+                )
             commit = self._repo.index.commit(msg)
             logger.info(f"commit: [{commit.hexsha}] {msg}")
```
gboutry, line 57:
```suggestion
    def validate(self, msg):
        with self._lock:
            yield self
            try:
                diffs = self._repo.index.diff("HEAD")
            except BadName as e:
                logger.debug(
                    f"error during diff: {e}. Probably because the repo is still empty."
                )
                diffs = []
            if len(diffs) == 0:
                raise EmptyCommit("The commit has no diff")
            commit = self._repo.index.commit(msg)
            logger.info(f"commit: [{commit.hexsha}] {msg}")
```

gboutry (changes requested):

gboutry (approved):

PR #155, opened 2022-05-03
Remove dag definition from tdp-lib, fix the cli to keep it working

### LINKED ISSUE #149: dag: move dag definition to tdp-collection
The dag definition (`tdp/components/*.yml`) are definitions that should be in the main collection.

They are heavily coupled with the main collection, and they should evolve together

gboutry:
fix #149

Requires:

- [x] TOSIT-IO/tdp-collection#250

With this PR, all commands using the dag must have the `collection-path` argument.


tdp/cli/utils.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2022 TOSIT.IO
+# SPDX-License-Identifier: Apache-2.0
+
+
+from tdp.core.dag import DAG_EXTENSION, DAG_FOLDER_NAME, Dag
+
+
+def create_dag_from_collection_path(collection_path):
```
rpignolet, line 8: Why not put this function as a static method inside `Dag` class and name it `from_collection()` ?

tdp/core/dag.py
```diff
@@ -46,21 +46,25 @@
...
         self._services_components = None
+        self.yaml_files = yaml_files
         self.playbooks_dir = playbooks_dir
 
-        if yaml_files is None:
-            yaml_files = list((Path(tdp.components.__file__).parent).glob("*.yml"))
-        self.yaml_files = yaml_files
+    @staticmethod
+    def from_collection(collection_path):
+        return Dag((collection_path / DAG_FOLDER_NAME).glob("*" + DAG_EXTENSION))
```
rpignolet, line 67: What do you think about using `Path` constructor in order to accept `str` for `collection_path` parameter ? Actually, `collection_path` must be a `Path` object. If `collection_path` is a `str`, the `/` operator will not work.

gboutry: done in 09bd0e0

rpignolet (approved):

leopaul36 (approved):

PR #159, opened 2022-05-10
cli: import libraries from extras that are needed later

### LINKED ISSUE #158: tdp cli doesn't work when the extras visualization are not installed
The cli fails with the following error:
```
2022-05-10 09:43:16,953 - DEBUG - tdp.<module> - Logger initialized
Traceback (most recent call last):
  File "<HOME_FOLDER>/tdp-lib/tdp/cli/commands/dag.py", line 16, in <module>
    import matplotlib
ModuleNotFoundError: No module named 'matplotlib'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<HOME_FOLDER>/.cache/pypoetry/virtualenvs/tdp-lib-HYsIegcK-py3.6/bin/tdp", line 2, in <module>
    from tdp.cli.__main__ import tdp
  File "<HOME_FOLDER>/tdp-lib/tdp/cli/__main__.py", line 7, in <module>
    from tdp.cli.commands.dag import dag
  File "<HOME_FOLDER>/tdp-lib/tdp/cli/commands/dag.py", line 21, in <module>
    ) from e
RuntimeError: You need to install the 'visualization' extras to be able to use the dag command. Run 'poetry install --extras visualization'
```

Only the command `tdp dag` should fail.

gboutry:
fix #158 

In this PR:
- `show` and the optional libraries from extras are imported later (at the start of the function), to prevent failing in the module loading phase.
- Improved output
```
2022-05-10 10:02:01,686 - DEBUG - tdp.<module> - Logger initialized
Error: You need to install the 'visualization' extras to be able to use the dag command. Run 'poetry install --extras visualization'
```

### DIFF
--- a/tdp/cli/commands/dag.py
+++ b/tdp/cli/commands/dag.py
@@ -9,15 +9,6 @@ import click
 import networkx as nx
 
 from tdp.core.dag import Dag
-from tdp.core.dag_dot import show
-
-try:
-    import matplotlib
-    import pydot
-except ImportError as e:
-    raise RuntimeError(
-        "You need to install the 'visualization' extras to be able to use the dag command. Run 'poetry install --extras visualization'"
-    ) from e
 
 SHORT_DAG_SUMMARY = "Compute and display a graph."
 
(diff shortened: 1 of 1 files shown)


rpignolet (approved):

gboutry:
Force-push is about fixing typo in commit message.

PR #191, opened 2022-07-18
Run Operation not in DAG

### LINKED ISSUE #190: Allow OperationRunner to run Operation which is not in the DAG
Currently, OperationRunner run only DAG Operation. An Operation not in the DAG can not be run.

rpignolet:
Fix #190 

Add `Collections` class to handle collections order.

`Collections` class is used to load `Operation` from multiple `Collection` object. `Collections` allow to have `Collection` which override `Operation` from a previous loaded `Collection`. The loading order is important.

Moreover `Collections` handle `Operation` inside the DAG and `Operation` not in the DAG.

### DIFF
--- a/tdp/cli/commands/dag.py
+++ b/tdp/cli/commands/dag.py
@@ -72,7 +72,7 @@ def dag(
     cluster,
 ):
     show = import_show()
-    dag = Dag.from_collections(collection_path)
+    dag = Dag(collection_path)
     graph = dag.graph
     if nodes:
         if pattern_format:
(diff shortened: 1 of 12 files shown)


rpignolet:
`Collections` extends `Mapping` in order to quacks like a `dict`

gboutry (approved):
Looks good to me.

Just an attention point that came up while using multiple collections:
It fails for tdp-extras because the service `airflow` has not been upgraded to the newer collection format (while `livy` has been).

We should create an issue in the `tdp-collection-extras` repo to keep track.

rpignolet:
Yes https://github.com/TOSIT-IO/tdp-collection-extras/issues/44 created.

Issue #140, opened 2022-04-27
Should tdp-lib register dry-run deploys in the database


leopaul36:
Currently, a dry run of a `deploy` command gets logged in the database. Should we keep doing this?

```
tdp deploy --targets hdfs_install --dry     
2022-04-27 11:16:54,176 - DEBUG - tdp.<module> - Logger initialized
2022-04-27 11:16:54,871 - WARNING - tdp.dag.validate - playbooks_dir is not defined, skip playbooks validations
Deploying to ['hdfs_install']
2022-04-27 11:16:54,969 - DEBUG - tdp.action_runner.run - Running action hadoop_client_install
2022-04-27 11:16:54,969 - INFO - tdp.ansible_executor.execute - [DRY MODE] Ansible command: ansible-playbook /home/leo/TOSIT-FR/ansible-tdp-roles/playbooks/hadoop_client_install.yml
[...]

tdp browse
2022-04-27 11:16:57,342 - DEBUG - tdp.<module> - Logger initialized
Deployments:
  id  sources    targets            filter    start                       end                         state    actions                                                  services
----  ---------  -----------------  --------  --------------------------  --------------------------  -------  -------------------------------------------------------  -----------
   1  None       hdfs_install       None      2022-04-27 11:16:54.969527  2022-04-27 11:16:54.980150  Success  hadoop_client_install,...,hdfs_ssl-tls_install           hadoop,hdfs

```

rpignolet:
TDP is not deployed when `deploy` in dry run mode so nothing should be recorded inside the database.

If I'm correct, currently it is the case to test the insertion in the database. Maybe @gboutry can confirm.

gboutry:
> TDP is not deployed when `deploy` in dry run mode so nothing should be recorded inside the database.
> 
> If I'm correct, currently it is the case to test the insertion in the database. Maybe @gboutry can confirm.

That's correct, dry mode only replaces the AnsibleExecutor with a mock executor (Which means, only ansible execution is mocked)

rpignolet:
Dry must not write to the database.

PR #192, opened 2022-07-19
Restart required

### LINKED ISSUE #24: Deduce target based on var files change (restart required)
When some var files change, deduce target nodes on the DAG to reach.

Need:
- [x] #102 
- [x] #107 
- [x] #108

rpignolet:
Fix #24 

Add restart required feature

Add a new CLI command `restart-required` which configure and restart needed services/components if a service/component DAG ancestors configuration is modified.

### DIFF
--- a/tdp/cli/commands/restart_required.py
+++ b/tdp/cli/commands/restart_required.py
@@ -1,0 +1,97 @@
+# Copyright 2022 TOSIT.IO
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+from pathlib import Path
+
+import click
+
+from tdp.cli.commands.queries import get_latest_success_service_version_query
+from tdp.cli.session import get_session_class
+from tdp.cli.utils import check_services_cleanliness, collection_paths
+from tdp.core.dag import Dag
+from tdp.core.runner.ansible_executor import AnsibleExecutor
+from tdp.core.runner.operation_runner import OperationRunner
+from tdp.core.service_manager import ServiceManager
+
+
+@click.command(short_help="Restart required TDP services")
+@click.option(
+    "--database-dsn",
+    envvar="TDP_DATABASE_DSN",
+    required=True,
+    type=str,
+    help=(
+        "Database Data Source Name, in sqlalchemy driver form "
+        "example: sqlite:////data/tdp.db or sqlite+pysqlite:////data/tdp.db. "
+        "You might need to install the relevant driver to your installation (such "
+        "as psycopg2 for postgresql)"
+    ),
+)
+@click.option(
+    "--collection-path",
+    envvar="TDP_COLLECTION_PATH",
+    required=True,
+    callback=collection_paths,  # transforms list of path into Collections
+    help=f"List of paths separated by your os' path separator ({os.pathsep})",
+)
+@click.option(
+    "--run-directory",
+    envvar="TDP_RUN_DIRECTORY",
+    type=Path,
+    help="Working directory where the executor is launched (`ansible-playbook` for Ansible)",
+    required=True,
+)
+@click.option(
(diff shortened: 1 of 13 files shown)


tdp/cli/commands/restart_required.py
```diff
@@ -0,0 +1,97 @@
...
+        operation_runner = OperationRunner(dag, ansible_executor, service_managers)
+
+        deployment = operation_runner.run_nodes(
+            sources=list(components_modified),
+            restart=True,
+        )
+
+        if not dry:
+            session.add(deployment)
+            session.commit()
```
gboutry, line 97: In the deploy command, even if it's dry, it's inserted in the database. Is there a reason this behavior is different in this command ?

tdp/cli/commands/restart_required.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2022 TOSIT.IO
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+from pathlib import Path
+
+import click
+
+from tdp.cli.commands.service_versions import get_latest_success_service_version
```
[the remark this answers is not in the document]
gboutry: Moreover, I think `get_latest_success_service_version` should become `get_latest_success_service_version_query`, not take the session as parameter and return a selectable.

tdp/cli/commands/restart_required.py
gboutry: ref #140

tdp/cli/commands/restart_required.py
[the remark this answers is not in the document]
rpignolet: Ok but in the actual implementation the `session` is used for `session.query()` how I replace this ?

tdp/cli/commands/restart_required.py
rpignolet: This is not the behavior that we want ? In `dry` mode it should not write in the database or I did not understand the #140.

tdp/cli/commands/restart_required.py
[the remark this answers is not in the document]
rpignolet: I move the function inside a `queries` file but I don't know how to transform `session.query` to `select`.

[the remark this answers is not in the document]
gboutry: You change the part `session.query(...).all()` to `select(...)` (without the all).

Then, when you use it:
`session.execute(the selectable object).all()`

tdp/cli/commands/restart_required.py
gboutry: Well the issue #40 questions it, but we didn't statuate on this yet. You're introducing new behavior before anything has been decided.

tdp/cli/commands/restart_required.py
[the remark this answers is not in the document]
rpignolet: Ok done.

tdp/cli/commands/restart_required.py
rpignolet: I deleted the `if`.

gboutry (approved):

tdp/cli/commands/restart_required.py
```diff
@@ -0,0 +1,97 @@
...
+                raise RuntimeError(
+                    f"Service '{service}' is deployed but the repository is missing."
+                )
+            for component_modified in service_managers[service].components_modified(
+                version
+            ):
+                components_modified.add(component_modified.name)
+
+        if len(components_modified) == 0:
+            click.echo("Nothing need to be restarted")
```
gboutry, line 87:
```suggestion
            click.echo("Nothing needs to be restarted")
```

rpignolet: done

PR #197, opened 2022-07-26
cli: remove operations in tdp browse deployments

### LINKED ISSUE #196: cli: 'tdp browse' launches several queries
In debug mode, we could observe that ```sqlalchemy``` launch ```(2 * n + 1)``` queries for ```n``` records of deployments.

kpgtek:
Fix #196 

The column about ```operations``` in ```deployments``` returned by ```tdp browse``` could be removed, because it can not contain a full list of operations, and this could be display with ```tdp browse <deployment_id>```.

By this way, we could thus reduce number of queries launched by ```tdp browse```.

### DIFF
--- a/tdp/cli/commands/browse.py
+++ b/tdp/cli/commands/browse.py
@@ -57,7 +57,9 @@ def browse(deployment_id, operation, database_dsn, limit, offset):
 
 
 def process_deployments_query(session_class, limit, offset):
-    headers = [key for key, _ in keyvalgen(DeploymentLog)]
+    headers = DeploymentLog.__table__.columns.keys() + [
+        str(DeploymentLog.services).split(".")[1]
+    ]
     query = select(DeploymentLog).order_by(DeploymentLog.id).limit(limit).offset(offset)
 
     with session_class() as session:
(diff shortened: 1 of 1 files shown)


tdp/cli/commands/browse.py
```diff
@@ -173,11 +177,6 @@ def custom_format(key, value):
                 return value[0] + ",...," + value[-1]
             else:
                 return ",".join(value)
-        elif key == "operations":
-            if len(value) > 2:
-                return value[0].operation + ",...," + value[-1].operation
-            else:
-                return ",".join(operation_log.operation for operation_log in value)
```
gboutry, line 180: If we keep the code as it was in `process_single_deployment_query`, we should keep this.

tdp/cli/commands/browse.py
```diff
@@ -75,7 +77,9 @@ def process_deployments_query(session_class, limit, offset):
 
 
 def process_single_deployment_query(session_class, deployment_id):
-    deployment_headers = [key for key, _ in keyvalgen(DeploymentLog)]
+    deployment_headers = DeploymentLog.__table__.columns.keys() + [
+        str(DeploymentLog.services).split(".")[1]
+    ]
```
gboutry, line 82: Why ? The operations logs for the deployments will be queried anyway. (at l122)
You should keep it as it was.

```suggestion
    deployment_headers = [key for key, _ in keyvalgen(DeploymentLog)]
```

gboutry (changes requested):

gboutry (approved):

rpignolet (approved):

PR #200, opened 2022-07-29
feat(models): make DeploymentLog explicitly inheritable

### LINKED ISSUE #199: DeploymentLog is not explicitly inheritable
If we want outside use of tdp-lib where we can enrich a DeploymentLog, we need to make it explicitly inheritable.

gboutry:
fix #199

This pr enables SQLAlchemy polymorphism to allow outside use (such as tdp-server), to cleanly inherit from this table and enrich it with more data.


rpignolet (approved):

gboutry:
Closing this PR

PR #221, opened 2022-09-21
fix: raise an error when parameters lead to an empty list of operations

### LINKED ISSUE #216: cli: "IndexError: list index out of range" with incorrect filter string
When the filter string is not correctly put for the option ```--filter``` of ```tdp deploy``` command, for example:
```
$ tdp deploy --targets zookeeper_init --filter 'zookeeper' --dry
```
I get the error message ```IndexError: list index out of range```.

gboutry:
fix #216

### DIFF
--- a/tdp/core/runner/operation_runner.py
+++ b/tdp/core/runner/operation_runner.py
@@ -66,6 +66,10 @@ class OperationIterator(Iterator):
         ]
 
 
+class EmptyOperationPlan(Exception):
+    pass
+
+
 class OperationPlan:
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


tdp/core/runner/operation_runner.py
```diff
@@ -120,7 +120,10 @@ def from_dag(
         else:
             str_filter = None
             filter_type = None
-
+        if len(operation_names) == 0:
+            raise ValueError(
```
rpignolet, line 124: I think we should not raise this generic exception, this code is part of a lib and the exception should be specific like `EmptyListOfOperations` but in case we do not raise exception we must decide if returning an `OperationPlan` without operation to run make sense. Other solution is to return `None` instead of an `OperationPlan`.

gboutry: It's done

rpignolet (approved):

PR #224, opened 2022-09-28
feat: refactor variables

### LINKED ISSUE #223: Refactor Variables to simplify variable handling
We should simplify Variables into a mutablemapping to be able to use it in python constructs more easily.

Moreover, the custom behavior for the methods `get` and `set` seems unnecessary

Implement a merge function to add merge hash

gboutry:
fix #223

Simplification of variables handling.

Returning deepcopy on get was a bad idea, and made handling vars very unpythonic.

`_flush_on_write` was utopic because it wasn't called on inner dict, or if the key's value was another kind of container (append on a list).

It's easier and more sound to flush when closing. Relying on caller either using the `Variables.open()` values into a context manager, or calling `close()` himself.

### DIFF
--- a/tdp/core/variables.py
+++ b/tdp/core/variables.py
@@ -1,14 +1,13 @@
 # Copyright 2022 TOSIT.IO
 # SPDX-License-Identifier: Apache-2.0
 
-import copy
 import logging
 import os
-from contextlib import contextmanager
-from functools import wraps
+from collections.abc import MutableMapping
 from weakref import proxy
 
 import yaml
+from ansible.utils.vars import merge_hash as _merge_hash
 
 try:
     from yaml import CDumper as Dumper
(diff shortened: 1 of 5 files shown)


rpignolet (approved):

tdp/core/variables.py
```diff
@@ -61,110 +64,45 @@ def __init__(self, content):
         """
         self._content = content
 
+    def copy(self):
+        return self._content.copy()
+
+    def merge(self, mapping):
```
rpignolet, line 70: This function return something ? Can you do something like `Variables(...).merge(...).merge(...)` to perform multiple merge or it is useless ?

gboutry: It would need to return `self`, I think this construct would be weird on a dict, I'd rather call sequentially multiple merge operation.

PR #226, opened 2022-10-03
feat: add possibility to add overrides to tdp init

gboutry:
Add option to add overrides to tdp initialization.

### DIFF
--- a/tdp/core/service_manager.py
+++ b/tdp/core/service_manager.py
@@ -60,7 +60,7 @@ class ServiceManager:
             return self.name + "_" + operation.component
         raise ValueError(f"Service {self.name} does not have a component {component}")
 
-    def initialize_variables(self):
+    def initialize_variables(self, override_folder=None):
 
         # dict with filename as key and a list of paths as value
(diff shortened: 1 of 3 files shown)


tdp/core/service_manager.py
```diff
@@ -101,12 +110,13 @@ def initialize_variables(self):
...
 
     @staticmethod
-    def initialize_service_managers(dag, services_directory):
+    def initialize_service_managers(dag, services_directory, override_folder=None):
         """get a dict of service managers, initialize all services if needed
 
         Args:
             dag (Dag): operations' DAG
             services_directory (Union[str, Path]): path of the tdp vars
+            override_folder (Optional(str, Path)): path of tdp vars overrides
```
rpignolet, line 119: `override_folder` is a `str` or a `Path` so `Optional[str | Path]` no ?

gboutry: Corrected documentation

rpignolet:
I test and it is working as expected for me.

PR #229, opened 2022-10-04
Refactor variable management

gboutry:
Refactor with breaking change on Variables management and Repository interface.

In this PR:
  - Repositories no longer instantiates Variables objet, neither do they open a path
  - dict `services_managers` becomes `ClusterVariables`, its only goal is to interact with services variables and not manage services
  - `ServiceManager` becomes `ServiceVariables`, this class goal is to allow operation on service variables, but not on `services`, there `manager` was not appropriate


tdp/core/variables/service_variables.py
```diff
@@ -0,0 +1,137 @@
...
+        if operations_filtered:
+            operation = operations_filtered[0]
+            return self.name + "_" + operation.component
+        raise ValueError(f"Service {self.name} does not have a component {component}")
+
+    def update_from_variables_folder(self, message, tdp_vars_overrides):
+        override_files = list(tdp_vars_overrides.glob("*" + YML_EXTENSION))
+        service_files_to_open = [override_file.name for override_file in override_files]
+        with self.open_var_files(
+            f"{self.name}: {message}", service_files_to_open
```
rpignolet, line 62: Prefix with the service name is useful ? Each repository is dedicated to a service so all commit will have a prefix with the same service name. I think we can remove this prefix.

tdp/core/variables/service_variables.py
```diff
@@ -0,0 +1,137 @@
...
+        ) as configurations:
+            for file in override_files:
+                logger.info(f"Initializing {self.name} with variables from {file}")
```
rpignolet, line 65: It is always an "init" ? When multiple `update_from_variables_folder` are called to add multiple collection and overrides, it will merge with existing and it is no longer an "init" no ?

rpignolet (changes requested):

tdp/core/variables/cluster_variables.py
```diff
@@ -0,0 +1,112 @@
...
+                except FileExistsError:
+                    if not service_tdp_vars.is_dir():
+                        raise ValueError(
+                            f"{service_tdp_vars.absolute()} should be a directory"
+                        )
+
+                if service in cluster_variables:
+                    service_variables = cluster_variables[service]
+                else:
+                    repo = GitRepository.init(service_tdp_vars)
```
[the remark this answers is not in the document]
gboutry: Done

tdp/core/variables/service_variables.py
gboutry: You're right, done.

tdp/core/variables/service_variables.py
gboutry: Done

rpignolet (approved):

gboutry:
Force push is rebase on master

PR #234, opened 2022-10-06
feat: refactor Dag method to return Operation instance instead of str

### LINKED ISSUE #233: Refactor Dag to return Operation instance instead of str for topological sort and filter
When running the `topological_sort` method of a `Dag` instance (via `get_operations` for example) it will return a list of `str` which represent a list of `Operation`. In many place, we must get an `Operation` instance from the operation name.

`networkx` can support adding custom data to graph nodes (via `graph.add_node`)

rpignolet:
Fix #233

### DIFF
--- a/tdp/core/runner/operation_runner.py
+++ b/tdp/core/runner/operation_runner.py
@@ -4,7 +4,6 @@
 import logging
 from collections.abc import Iterator
 from datetime import datetime
-from operator import itemgetter
 
 from tdp.core.models.deployment_log import DeploymentLog, FilterTypeEnum
 from tdp.core.models.operation_log import OperationLog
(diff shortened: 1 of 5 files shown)


tdp/core/runner/operation_runner.py
```diff
@@ -110,26 +110,20 @@ def get_deployment_args(self):
...
-                operation_names, filter_expression
-            )
+            operations = dag.filter_operations_glob(operations, filter_expression)
             str_filter = filter_expression
             filter_type = FilterTypeEnum.GLOB.value
         else:
             str_filter = None
             filter_type = None
-        # Converting operations names into their actual Operation object
-        operations = itemgetter(*operation_names)(dag.operations)
```
gboutry, line 130: Can you remove the import associated with `itemgetter` ?

rpignolet: done

rpignolet: A linter is missing for checking unused import :)

gboutry (approved):

PR #239, opened 2022-10-06
feat: add resume mode to deploy

### LINKED ISSUE #225: Add a resume mode for tdp-lib
We could use to have a `--resume` mode in `tdp deploy`.

Without parameter, it would resume the last deploy action using the same DAG, skipping the successful nodes. With a parameter it would resume a given deployment id.

An attempt to resume a Successful deployment should raise an exception.

leopaul36:
Fix #225 

A few remarks:
- I wanted to put a default value for `--resume` like `latest` which would be interpreted as the latest deployment but I did not know how to do it since `get_deployments` takes a `limit` parameter 
- I chose to make `--resume` and `--sources` incompatible as I think it does not make sense to have both
- I am not sure if `click.BadParameter` is a relevant exception for the `Nothing to resume...` part

### DIFF
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -8,6 +8,7 @@ import click
 
 from tdp.cli.session import get_session_class
 from tdp.cli.utils import check_services_cleanliness, collection_paths_to_collections
+from tdp.cli.commands.browse import get_deployment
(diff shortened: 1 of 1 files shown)


leopaul36:
New PR: https://github.com/TOSIT-IO/tdp-lib/pull/244

PR #244, opened 2022-10-27
feat: add resume mode to deploy

leopaul36:
#### Which issue(s) this PR fixes:

Fixes #225 

#### Additional comments:

To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0) and that you give permission to [TOSIT](https://www.tosit.io/), you have to acknowledge this by using the following check-box.

 - [X] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0).
 - [X] I hereby agree to grant [TOSIT](https://www.tosit.io/) a copyright license to use my contributions.

PR #256, opened 2022-11-08
feat: add resume option

### LINKED ISSUE #225: Add a resume mode for tdp-lib

leopaul36:

Fixes #225

Main features:
- Add `--resume` option
- Move `get_deployment` to `queries.py`
- Introduced new Enum `DeploymentTypeEnum`
- Also applied the non resolved comments from https://github.com/TOSIT-IO/tdp-lib/pull/244

#### Additional comments:

### DIFF
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -68,7 +68,7 @@ from tdp.core.variables import ClusterVariables
     "--resume",
     type=str,
     metavar="dep_id",
-    help="Resume a previous deployment, cannot be used with --sources or --targets",
+    help="Resume a previous deployment, cannot be used with --sources, --targets or --filter",
 )
 def deploy(
     sources,
@@ -109,9 +109,9 @@ def deploy(
             collections, ansible_executor, cluster_variables
         )
         if resume:
-            if sources or targets:
+            if sources or targets or filter:
                 raise click.BadParameter(
-                    f"sources or targets can not be used with resume"
+                    f"sources, targets or filter can not be used with resume"
                 )
             resumed_deployment = get_deployment(session_class, resume)
             if (
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -8,7 +8,9 @@ import click
 
 from tdp.cli.session import get_session_class
 from tdp.cli.utils import check_services_cleanliness, collection_paths_to_collections
+from tdp.cli.commands.queries import get_deployment
 from tdp.core.dag import Dag
+from tdp.core.models.state_enum import StateEnum
 from tdp.core.runner import AnsibleExecutor, DeploymentPlan, DeploymentRunner
 from tdp.core.variables import ClusterVariables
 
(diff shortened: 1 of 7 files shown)


tdp/cli/commands/deploy.py
```diff
@@ -62,6 +64,12 @@
 )
 @click.option("--filter", type=str, help="Glob on list name")
 @click.option("--dry", is_flag=True, help="Execute dag without running any operation")
+@click.option(
+    "--resume",
+    type=str,
+    metavar="dep_id",
+    help="Resume a previous deployment, cannot be used with --sources or --targets",
```
rpignolet, line 71: `--filter` is allowed ? If the "resumed deployment" had a `filter`, the `--filter` will be overridden. I think `--filter` cannot be used with `resume` too.

tdp/cli/commands/deploy.py
```diff
@@ -99,16 +108,62 @@ def deploy(
...
+                ]
+                # print(original_operations)
+                succeeded_operations = [
+                    operation.operation
+                    for operation in resumed_deployment.operations
+                    if operation.state == StateEnum.SUCCESS
+                ]
+                # print(succeeded_operations)
+                remaining_operations_list = list(
+                    set(original_operations) - set(succeeded_operations)
```
rpignolet, line 154: This logic can not be moved in `DeploymentPlan` class ? We can add a method in `DeploymentPlan` which take a list of operations and change the plan with only remaining operations. Or maybe the method can take a `DeploymentLog` instance.

rpignolet (changes requested):

tdp/cli/commands/deploy.py
```diff
@@ -62,6 +64,12 @@
 )
 @click.option("--filter", type=str, help="Glob on list name")
 @click.option("--dry", is_flag=True, help="Execute dag without running any operation")
+@click.option(
```
gboutry, line 67: Does this mean we need to pass the deployment id anytime we want to resume ? Can't we resume automatically if previous deployment was a failure ?

leopaul36: We can do this, would a `get_last_deployment` with a query like `select(DeploymentLog).order_by(DeploymentLog.id.desc()).first()` be fine?

gboutry: Yes

tdp/core/runner/deployment_plan.py
```diff
@@ -51,6 +52,14 @@ def from_dag(
     def from_operations(operations):
         deployment_args = dict(
             targets=[operation.name for operation in operations],
-            using_dag=False,
+            deployment_type=DeploymentTypeEnum.OPERATIONS,
+        )
+        return DeploymentPlan(operations, deployment_args)
+
+    @staticmethod
```
gboutry, line 59: This should take a DeploymentLog as input (not operations), and should fail if the deploymentlog is in success

leopaul36:
Closed in favor of https://github.com/TOSIT-IO/tdp-lib/pull/258

PR #257, opened 2022-11-08
203 store deployed component version instead of service version in rdbms

### LINKED ISSUE #203: Store deployed component version instead of service version in RDBMS
With #19, deployed service version is stored in RDBMS. When deploying a service, the version of this service is stored in the RDBMS regardless of the component of the service deployed.

We must store the version of each deployed component, and only consider it deployed if the component's `config` action is successfully completed AND the component's `start` or `restart` action is successfully completed with "changed".

gboutry:

Fixes #203

#### Additional comments:


tdp/cli/commands/browse.py
```diff
@@ -139,11 +143,11 @@ def print_formatted_deployment(deployment_log):
     click.echo("Sources:\n  " + tabulate({"source": sources}, headers="keys"))
     click.echo("Targets:\n  " + tabulate({"target": targets}, headers="keys"))
     click.echo(
-        "Services:\n"
+        "Service Component logss:\n"
```
rpignolet, line 146: Extra `s`.
```suggestion
        "Service Component logs:\n"
```

tdp/cli/commands/browse.py
```diff
@@ -162,14 +166,14 @@ def print_formatted_deployment(deployment_log):
 
 def print_formatted_operation_log(operation_log):
     headers = OperationLog.__table__.columns.keys()
-    service_headers = ServiceLog.__table__.columns.keys()
-
+    service_headers = ServiceComponentLog.__table__.columns.keys()
+    # TODO: this outputs Service and ServiceComponent version
```
rpignolet, line 170: Can you rewrite the `TODO` ? I don't understand what should be done.

tdp/cli/commands/deploy.py
```diff
@@ -113,14 +113,16 @@ def deploy(
...
-            for operation in deployment_iterator:
+            for _ in deployment_iterator:
                 pass
         else:
             session.add(deployment_iterator.log)
             # insert pending deployment log
             session.commit()
-            for operation in deployment_iterator:
-                session.add(operation)
+            for operation_log, service_log in deployment_iterator:
```
rpignolet, line 122: It is a `service_log` ? Not a `service_component_log` ?

tdp/cli/commands/browse.py
gboutry: Fixed

tdp/cli/commands/browse.py
gboutry: I added information to the todo

tdp/cli/commands/deploy.py
gboutry: Fixed

gboutry:
Force push is adding condition in `and service_component.log is None` when emitting service component log for edge cases when you have the following operation `config, start, restart` or `config,start,config,start` on the same component. In any case, it should only emit 1 service_component_log.

This should not be possible with operations from a dag, but it is when passing your own operation list.

gboutry:
Converting back to draft, `tdp restart-required` does not work, it should be upgraded to new service component logs

gboutry:
@rpignolet Reopened with naive resolution to restart-required, the current version is not the most efficient but should work. (At worst case every component of a service is restarted even if it shouldn't).

I would prefer to tackle this problem in a separate PR.

tdp/core/dag.py
```diff
@@ -145,14 +145,17 @@ def custom_key(node):
...
             return f"{operation_priority:02d}_{node}"
 
-        def to_restart(node):
+        def to_node(node):
             if restart:
-                return node.replace("_start", "_restart")
-            return node
+                node_replaced = node.replace("_start", "_restart")
+                node_restart = self.collections.operations.get(node_replaced)
+                if node_restart is not None:
```
rpignolet, line 152: I do not understand why you add this condition ? If the `restart` action does not exit the code must raise an exception. If I understand the new code, the `start` action will be used instead of the `restart` if the `restart` does not exists.

gboutry: Noop will also go through this part, and noop starts don't have a restart playbook (since they don't even have a start playbook)

rpignolet: We can get the actual operation with `self.collections.operations.get(node)` and check if it is a `noop` ? If the operation is `noop` we do not perform the `node.replace("_start", "_restart")`. I think we must keep the error when a `restart` action is missing because without `restart` the new configuration will not be applied.

gboutry: If start operation was a noop, a noop restart operation will be created instead of returning the original start operation, since the caller expects to only get restart operation as a result

rpignolet (approved):

tdp/core/dag.py
rpignolet: ok

PR #274, opened 2022-11-28
fix(model): introduce surrogate key in ServiceComponentLog

### LINKED ISSUE #273: service_component_log introduce surrogate primary key
In recent updates, I changed the schema of the table `service_log` into `service_component_log`, where the `component` attribute was part of the primary key, but `NULLABLE`, this is legal in SQLite, but not in PostgreSQL.

We need to introduce a surrogate primary key in the table to be able to use PostgreSQL.

gboutry:

Fixes #273

#### Additional comments:

Having part of a composite primary key nullable is forbidden in Postgresql, therefore we needed to introduce a surrogate key to become the new primary key.

#### Agreements

### DIFF
--- a/tdp/core/models/service_component_log.py
+++ b/tdp/core/models/service_component_log.py
@@ -1,7 +1,7 @@
 # Copyright 2022 TOSIT.IO
 # SPDX-License-Identifier: Apache-2.0
 
-from sqlalchemy import Column, ForeignKey, Integer, String
+from sqlalchemy import Column, ForeignKey, Integer, String, UniqueConstraint
 from sqlalchemy.orm import relationship
 
 from tdp.core.models.base import Base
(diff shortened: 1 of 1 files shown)


rpignolet (approved):

tdp/core/models/service_component_log.py
```diff
@@ -12,13 +12,19 @@
...
     version = Column(String(length=VERSION_MAX_LENGTH), nullable=False)
 
     deployment = relationship("DeploymentLog", back_populates="service_components")
+
+    __table_args__ = (
+        UniqueConstraint(
+            "deployment_id",
+            "service",
+            "component",
+            name="_unique_service_component_constraint",
```
kpgtek, line 28: I think it is better to remove the first '_' character for ```_unique_service_component_constraint``` to avoid unexpected situation.
It seems that the name is not used by SQLite for the index. But it is used by PostgreSQL as name for the index of that constraint:
```
Indexes:
    "service_component_log_pkey" PRIMARY KEY, btree (id)
    "_unique_service_component_constraint" UNIQUE CONSTRAINT, btree (deployment_id, service, component)
```

If the name is not specified, the default Postgres index name for the constraint will be: 
```
<table>_<column1>_<column2>_<column3>_<...>_key
```
For our example, it will be:
```
service_component_log_deployment_id_service_component_key
```

gboutry: I removed the name to let the SGBD setup its own name

kpgtek (approved):

PR #290, opened 2022-12-26
feat: enable schema validation

### LINKED ISSUE #289: Enable schema validation in tdp-lib
Following https://github.com/TOSIT-IO/tdp-collection/pull/609, we need tdp-lib to be able to validate the schema, and expose the schema through its api.

gboutry:

Fixes #289

#### Additional comments

This PR adds schema validation to tdp-lib, and allows a collection to expose it's schema definitions.

Collections can extend an existing json schema, this will be treated as additional constraints. 
This means incompatibilities between 2 definitions of a same schema will raise an exception in validation stage.

Schema definition is allowed at service level, not component for the moment.

Requires:
- [x] https://github.com/TOSIT-IO/tdp-collection/pull/609

#### Agreements


tdp/core/collection.py
```diff
@@ -3,10 +3,14 @@
 
 from pathlib import Path
 
+from networkx.readwrite.json_graph.jit import json
```
rpignolet, line 6: Why not use `import json` ?

tdp/core/variables/cluster_variables.py
```diff
@@ -27,7 +28,11 @@ def __iter__(self):
 
     @staticmethod
     def initialize_cluster_variables(
-        collections, tdp_vars, override_folder=None, repository_class=GitRepository
+        collections,
+        tdp_vars,
+        override_folder=None,
+        repository_class=GitRepository,
+        validate=False,
```
rpignolet, line 35: Why disable validation ?

gboutry: To make it opt-in, there might be use case where we define the schema wrongly, and we shouldn't prevent initialization is a tdp-lib user **realy** wants to initialize

tdp/core/collection.py
gboutry: Oups, automatic import from my ide, it's fixed

tdp/core/variables/cluster_variables.py
rpignolet: Yes, and this is a default value, the CLI call it with `True` by default.

rpignolet (approved):

leopaul36 (approved):

PR #302, opened 2023-03-06
feat(cli/init): allow multiple overrides paths

### LINKED ISSUE #301: Ability to provide multiple paths to the 'tdp init --overrides' option
This is an improvement request following the https://github.com/TOSIT-IO/tdp-collection/issues/713's discussion with @rpignolet. It appeared that adding the ability to provide multiple paths to the `--overrides` option of the `init` command would allow users to better manage their own environments.

Let's imagine I can deploy TDP clusters in multiple environments, say 3 environments, physically located in 2 different technical environments :
- platform1
  - dev
  - staging
- platform2
  - prod

The idea is to never define twice the same thing. With the following structure, I could define variables in `platform2/` folder that I want to be common to `staging` and `dev` environments, and in the `dev/` folder only, `dev` specific variables :
```
my_vars/
├── dev
│   └── hadoop
│       └── hadoop.yml
├── platform1
│   └── hadoop
│       └── hadoop.yml
├── platform2
│   └── hadoop
│       └── hadoop.yml
├── prod
│   └── hadoop
│       └── hadoop.yml
└── staging
    └── hadoop
        └── hadoop.yml
```

With that structure in mind, I could do something like this :
```bash
# init overrides with multiple path
tdp init --overrides my_vars/platform1:my_vars/dev
# or even
tdp init --overrides myvars/platform1 --overrides my_vars/dev
```
And expect my TDP_VARS folder to be populated with the result of the merge of `tdp_vars_defaults/` from collections + `my_vars/platform1/` + `my_vars/dev/`. The order of the paths would be important, as the last path provided would win precedence over everything.

jusstol:

Fixes #301 

#### Additional comments
Here is a proposal for the linked issue to provide multiple paths. Those changes rely on Click's `multiple=True` option for command line options.
##### Usage
There are two ways to provide multiple paths :
- multiple `--overrides` from CLI :
```bash
tdp init --overrides my_vars/platform1 --overrides my_vars/dev
```
- multiple overrides paths from `TDP_OVERRIDES` environment variable :
```bash
export TDP_OVERRIDES=my_vars/platform1:my_vars/dev
tdp init
```
Both give the same output
```
2023-03-06 17:34:43,936 - INFO - tdp.cluster_variables.initialize_cluster_variables - hadoop is already initialized at 5f0a6bbb62d169c95d8c8079a989fd4200bcd04c
2023-03-06 17:34:43,942 - INFO - tdp.service_variables.update_from_variables_folder - Updating hadoop with variables from /my_env/my_vars/platform1/hadoop/hadoop.yml
2023-03-06 17:34:43,950 - DEBUG - tdp.git_repository.add_for_validation - hadoop.yml staged
2023-03-06 17:34:43,969 - INFO - tdp.git_repository.validate - commit: [8698ef55295c99d278e35420f8bd3888c5f75314] add variables from overrides_option_0
2023-03-06 17:34:43,969 - INFO - tdp.cluster_variables.initialize_cluster_variables - hadoop is already initialized at 8698ef55295c99d278e35420f8bd3888c5f75314
2023-03-06 17:34:43,975 - INFO - tdp.service_variables.update_from_variables_folder - Updating hadoop with variables from /my_env/my_vars/dev/hadoop/hadoop.yml
2023-03-06 17:34:43,980 - DEBUG - tdp.git_repository.add_for_validation - hadoop.yml staged
2023-03-06 17:34:43,999 - INFO - tdp.git_repository.validate - commit: [56f6512e8cc0ad7b87ebb9b38dc8fa3d2dd7ab56] add variables from overrides_option_1
```
##### Tests
- providing one path still works
- if both `--overrides` from CLI and environment variable `TDP_OVERRIDES` are used at the same time, only `--overrides` CLI options will be used and `TDP_OVERRIDES` will be ignored (`click`'s behavior)
- providing tdp overrides paths (NO `--overrides` CLI option and NO `TDP_OVERRIDES`) still works
- `poetry run pytest tdp` passed

#### Agreements

### DIFF
--- a/tdp/cli/commands/init.py
+++ b/tdp/cli/commands/init.py
@@ -16,7 +16,8 @@ from tdp.core.variables import ClusterVariables
     envvar="TDP_OVERRIDES",
     required=False,
     type=click.Path(exists=True, resolve_path=True, path_type=Path),
-    help="Path to tdp vars overrides",
+    multiple=True,
+    help="Path to tdp vars overrides. Can be used multiple times. Last one takes precedence.",
 )
 @collections
 @database_dsn
--- a/tdp/core/variables/cluster_variables.py
+++ b/tdp/core/variables/cluster_variables.py
@@ -30,7 +30,7 @@ class ClusterVariables(Mapping):
     def initialize_cluster_variables(
         collections,
         tdp_vars,
-        override_folder=None,
+        override_folders=None,
         repository_class=GitRepository,
         validate=False,
     ):
@@ -39,11 +39,14 @@ class ClusterVariables(Mapping):
         Args:
             collections (Collections): instance of collections
             tdp_vars (Union[str, Path]): path to the tdp vars
-            override_folder (Optional[str | Path]): path of tdp vars overrides
+            override_folders (Iterable[str | Path]): list of path(s) of tdp vars overrides
 
         Returns:
             ClusterVariables: mapping of service with their ServiceVariables instance
         """
+        if override_folders is None:
+            override_folders = []
+        
         tdp_vars = Path(tdp_vars)
 
         cluster_variables = {}
@@ -52,9 +55,12 @@ class ClusterVariables(Mapping):
             (collection_name, collection.default_vars_directory.iterdir())
             for collection_name, collection in collections.items()
         ]
-        if override_folder:
+
+        for i, override_folder in enumerate(override_folders):
             override_folder = Path(override_folder)
-            collections_and_overrides.append(("overrides", override_folder.iterdir()))
+            collections_and_overrides.append(
+                (f"overrides_path_{i}", override_folder.iterdir())
+            )
 
         # If the service was already initialized, we do not touch it
         services_initialized_by_this_function = set()


tdp/core/variables/cluster_variables.py
```diff
@@ -39,7 +39,7 @@ def initialize_cluster_variables(
         Args:
             collections (Collections): instance of collections
             tdp_vars (Union[str, Path]): path to the tdp vars
-            override_folder (Optional[str | Path]): path of tdp vars overrides
+            override_folders (Optional[tuple]): tuple of path(s) of tdp vars overrides
```
rpignolet, line 42: It's more common to have lists than tuples.
```suggestion
            override_folders (Optional[List[str | Path]]): list of path(s) of tdp vars overrides
```

tdp/core/variables/cluster_variables.py
```diff
@@ -52,9 +52,11 @@ def initialize_cluster_variables(
             (collection_name, collection.default_vars_directory.iterdir())
             for collection_name, collection in collections.items()
         ]
-        if override_folder:
-            override_folder = Path(override_folder)
-            collections_and_overrides.append(("overrides", override_folder.iterdir()))
+
+        for i, override_folder in enumerate(override_folders):
+            collections_and_overrides.append(
```
rpignolet, line 57: We accept folder as `str` and `Path` for `override_folder`, if you have a `str` the `.iterdir()` will fail, you must keep the `override_folder = Path(override_folder)`.

tdp/cli/commands/init.py
```diff
@@ -16,7 +16,8 @@
     envvar="TDP_OVERRIDES",
     required=False,
     type=click.Path(exists=True, resolve_path=True, path_type=Path),
-    help="Path to tdp vars overrides",
+    multiple=True,
```
rpignolet, line 19: I think we should have the same behavior as `TDP_COLLECTION_PATH`, use the OS path separator to have multiple path in order to specify multiple path with `TDP_OVERRIDES` env variable.

rpignolet (changes requested):

tdp/cli/commands/init.py
jusstol: I showed in the description of this PR that with this solution, `TDP_OVERRIDES` can accept multiple paths and behaves exactly like `TDP_COLLECTION_PATH` with OS separator, as it is natively supported by Click. Here is what Click's documentation states :
> The default implementation for all types is to split on whitespace. The exceptions to this rule are the [File](https://click.palletsprojects.com/en/8.1.x/api/#click.File) and [Path](https://click.palletsprojects.com/en/8.1.x/api/#click.Path) types which both split according to the operating system’s path splitting rules. On Unix systems like Linux and OS X, the splitting happens for those on every colon (:), and for Windows, on every semicolon (;).

See documentation and example here : https://click.palletsprojects.com/en/8.1.x/options/#multiple-values-from-environment-values.

I thought it was the simplest way to allow multiple paths, letting Click do the platform independent parsing. In fact,
 `TDP_COLLECTION_PATH` (`--collection-path`) could benefit from this too.

Let me know what you think, and I'll make change to this PR accordingly. Thanks for your review.

rpignolet: Ok to keep `multiple=true` if this the standard way with Click. I open #303 to change `--collection-path` behavior to match the new one.

tdp/core/variables/cluster_variables.py
jusstol: I changed the type hint in this docstring to 'tuple' and the default value to en empty tuple for two reasons :
- unlike lists, tuples are immutable and, the order of the paths here is important
- when using `multiple=True`, the result of Click's parsing always gives a tuple of Paths (not a list) even if there was only one Path provided

Maybe we can change it to accept both `List[str | Path]` and `Tuple[str | Path]` as long as it's an iterable of `str` and/or `Path` objects ?

rpignolet: > * unlike lists, tuples are immutable and, the order of the paths here is important

Lists are ordered like tuples, the immutable behavior is not needed in this function.

> Maybe we can change it to accept both `List[str | Path]` and `Tuple[str | Path]` as long as it's an iterable of `str` and/or `Path` objects ?

The function need an iterable of `str` or `Path` yes, you can use `Iterable[str | Path]` if you want.

tdp/core/variables/cluster_variables.py
```diff
@@ -30,7 +30,7 @@ def __iter__(self):
     def initialize_cluster_variables(
         collections,
         tdp_vars,
-        override_folder=None,
+        override_folders=[],
```
rpignolet, line 33: Ok if you want to put `()` since it is immutable but you must avoid mutable type in default see https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/. Use `None` and in the function you can do:
``` python
if override_folders is None:
  override_folders = []
```

tdp/core/variables/cluster_variables.py
```diff
@@ -39,11 +39,14 @@ def initialize_cluster_variables(
         Args:
             collections (Collections): instance of collections
             tdp_vars (Union[str, Path]): path to the tdp vars
-            override_folder (Optional[str | Path]): path of tdp vars overrides
+            override_folders (Iterable[str | Path]): list of path(s) of tdp vars overrides
 
         Returns:
             ClusterVariables: mapping of service with their ServiceVariables instance
         """
+        if not isinstance(override_folders, (list, tuple)):
```
rpignolet, line 47: You verify the type of `override_folders`, we do not verify the type of every parameters so why verify the type for this variable ? If `override_folders` is not an `Iterable` it will fail in the `for` loop below this is fine for me.

rpignolet (changes requested):

tdp/core/variables/cluster_variables.py
jusstol: Since the `str` type is an iterable object, if `override_folders` is equal to `'my_vars/platform1'` and not `['my_vars/platform1']`, the first `for` loop will not fail because it will loop on every character of the string, and also because the `.iterdir()` method creates a generator without checking if the Path exists at that moment.

But in the end you are right, it will eventually fail later on at line 70 in the following `for` loop when trying to iterate over the "one character" path that probably won't exist (`for path in folder_iterator:`), giving a `FileNotFoundError`.

tdp/core/variables/cluster_variables.py
```diff
@@ -52,9 +55,12 @@ def initialize_cluster_variables(
             (collection_name, collection.default_vars_directory.iterdir())
             for collection_name, collection in collections.items()
         ]
-        if override_folder:
+
+        for i, override_folder in enumerate(override_folders):
             override_folder = Path(override_folder)
-            collections_and_overrides.append(("overrides", override_folder.iterdir()))
+            collections_and_overrides.append(
+                (f"overrides_option_{i}", override_folder.iterdir())
```
rpignolet, line 62: Out of curiosity, why did you use `overrides option_{i}` and not `overrides_{i}`. I don't really understand what the word "option" means.

jusstol: This was to reflect the fact that "--overrides" is a command line option, but I agree that's not the best idea as the paths can also come from the environment variable. We can call it "overrides_path_{i}" or just "overrides_{i}" as you suggested.

rpignolet: The code in this function is not related to the CLI, for example, this function is called inside `tdp-lib` CLI and `tdp-server` too. `overrides_path_{i}` is great.

jusstol: Changed it to `overrides_path_{i}`. Do you want me to squash my 4 commits into one, or the merge will take care of it ?

mdrutel (approved):
Use cases tested :
- one ```--overrides``` options => OK
- multiple ```--overrides``` options => OK
- ```--overrides``` and ```TDP_OVERRIDES``` => OK
- ```TDP_OVERRIDES``` => OK

jusstol:
Thank you for accepting this change!

PR #325, opened 2023-06-09
fix: prevent init fail when empty override

### LINKED ISSUE #324: `tdp init --overiddes` fails if one of the overrides file is empty
Reproduce:
```
tdp init --overrides tdp_vars_overrides
```

If one of the file in the override is empty (e.g. `tdp_vars_overrides/hive/hive.yml`) init crashes with error

```
raise EmptyCommit(
tdp.core.repository.repository.EmptyCommit: validating these changes would produce no difference
```

PaulFarault:

Fixes #324 

#### Agreements

### DIFF
--- a/tdp/core/variables/cluster_variables.py
+++ b/tdp/core/variables/cluster_variables.py
@@ -103,7 +103,7 @@ class ClusterVariables(Mapping):
                         )
                     except EmptyCommit:
                         logger.warning(
-                            f"override file {service_tdp_vars.absolute()} is empty, no commit has been made"
+                            f"override file {service_tdp_vars.absolute()} will not cause any change, no commit has been made"
                         )
 
         cluster_variables = ClusterVariables(cluster_variables)
--- a/tdp/core/variables/cluster_variables.py
+++ b/tdp/core/variables/cluster_variables.py
@@ -6,7 +6,7 @@ from collections.abc import Mapping
 from pathlib import Path
 
 from tdp.core.repository.git_repository import GitRepository
-from tdp.core.repository.repository import NoVersionYet
+from tdp.core.repository.repository import NoVersionYet, EmptyCommit
 
 from .service_variables import ServiceVariables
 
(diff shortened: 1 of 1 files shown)


tdp/core/variables/cluster_variables.py
```diff
@@ -97,9 +97,14 @@ def initialize_cluster_variables(
...
                 if service in services_initialized_by_this_function:
-                    service_variables.update_from_variables_folder(
-                        "add variables from " + collection_name, path
-                    )
+                    try:
+                        service_variables.update_from_variables_folder(
+                            "add variables from " + collection_name, path
+                        )
+                    except EmptyCommit:
+                        logger.warning(
```
rpignolet, line 105: Why put a logger here ? Why `warning` ? Why you do not raise the exception to the caller ? I do not understand the logic of this PR...

PaulFarault: > Why put a logger here ? [...] Why you do not raise the exception to the caller ?

Because in my understanding it's a non blocking exception.

Furthermore, I took inspiration from the rest of the method: https://github.com/TOSIT-IO/tdp-lib/blob/97add22a3ca7bf4185a72d17f20ff24f250ba60f/tdp/core/variables/cluster_variables.py#L66-L107

- > Why warning ?

Because that's what was suggested in #324. I could put a log if needed.

rpignolet: The problem is that you interpret an `except EmptyCommit` saying that the override file is empty but that is not what the `EmptyCommit` means. `EmptyCommit` means that there is no change following the merge of variables and this can happen if the override file is filled with variables identical to what is in the repository currently.

PACordonnier: This is true and the Warning log mesage is not appropriate but the issue still remains. The `EmptyCommit`  exception during validation should not interrupt the `init`

rpignolet: Yes the issue should be resolved but this changes should not be merged, this is not the correct way to fix it.

PaulFarault: Ok I understand now. Well, we may still want to keep this except in the case where the override file doesn't modify anything?

PaulFarault: As discussed, I will rephrase the message in a new commit!

PaulFarault: Done, I'll squash at merge.

PACordonnier (approved):

rpignolet (approved):

PR #342, opened 2023-06-30
[doc] Minor fix docstrings

sergkudinov:

None

#### Agreements

### DIFF
--- a/tdp/core/deployment/deployment_plan.py
+++ b/tdp/core/deployment/deployment_plan.py
@@ -49,7 +49,7 @@ class DeploymentPlan:
 
     Attributes:
         operations: List of operations to perform.
-        deployment_args: Deployment arguments.
+        deployment_log: Deployment information
     """
 
     def __init__(self, operations: List[Operation], deployment_log: DeploymentLog):


PaulFarault (approved):

jpsavy (commented):
One might want to rename the deployment_log attribute to _deployment_log so that it has a name consistent with the other static attributes (such as: _operations or _run_operation) of the DeploymentIterator class.

jpsavy (commented):
The deployment_log attribute could be prefixed with "_" in another PR: like 'Prefix deployment_log attribute with "_"'

rpignolet (approved):

PaulFarault:
@jpsavy The `_deployment_log` attribute has a getter and a setter (through the `@property` decorator), which means that we can access with `DeploymentPlan.deployment_log`.

Hence, it's not really an attribute but it has the same behavior so I think that's ok.

The true question may be "Why do we use a `@property` here?", it doesn't seem useful in this case. Any thoughts @rpignolet?

However, it is not related to this PR so we can merge it.

jpsavy (approved):

rpignolet:
@sergkudinov Please, resolve conflicts.

PR #350, opened 2023-07-05
test: add cli tests

### LINKED ISSUE #332: [test] Add test for the CLI
The CLI doesn't have any tests. We should at least add some to check if no error are thrown.

No need to check for the output for now as some commands can be tedious to test (i.e. `tdp browse`) and as we may want to refactor some part of the CLI with the server later.

sergkudinov:

Fixes #332

#### Agreements


sergkudinov:
Created a dedicated issu for testing `tdp dag` - #351

tdp/cli/commands/plan/test_reconfigure.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2022 TOSIT.IO
+# SPDX-License-Identifier: Apache-2.0
+
+from click.testing import CliRunner
+from .reconfigure import reconfigure
+from ..init import init
+
+
+def test_tdp_plan_reconfigure(collection_path, database_dsn_path, vars, tmp_path):
```
rpignolet, line 9: `tmp_path` is not used. Same for many tests.

tdp/cli/commands/plan/test_dag.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_tdp_plan_dag(collection_path, database_dsn_path, vars):
+    args = [
+        "--collection-path",
+        collection_path,
+        "--database-dsn",
+        database_dsn_path,
+        "--vars",
+        vars,
+    ]
```
rpignolet, line 17: Can you make a utility function that takes as arguments `collection_path`, `database_dsn_path`, etc. who generates this array so that the option names are only written in one place?

tdp/cli/commands/plan/test_dag.py
```diff
@@ -0,0 +1,21 @@
...
+    runner = CliRunner()
+    runner.invoke(init, args)
```
rpignolet, line 19: Why don't you check the return code of the `init` command?

rpignolet (changes requested):

tdp/cli/commands/plan/test_dag.py
sergkudinov: I tried, but I found it not very usefull. Many tests are taking different args like `database_dsn_path` or `database_dsn_path` or even without. I think it is more clear to keep it like this.

tdp/cli/commands/plan/test_dag.py
sergkudinov: becuse this test is testing `dag`, not `init`.  For `init` I have a dedicated test.

tdp/cli/commands/plan/test_reconfigure.py
sergkudinov: fixed

tdp/cli/commands/plan/test_dag.py
rpignolet: Yes `init` has its own test but if the `init` is in error, it doesn't make sense to continue the test and run the next command. In the error output it will show us that the following command is in error when it is the `init` which is in question.

rpignolet: This is not a unit test, you perform a `init` and a second command. If you want to do just the second command and do a real unit test you must mock the `init` command and not call the real one.

sergkudinov: agree, it is not a unit test. Ok, I'll change

sergkudinov:
@rpignolet 
ready fo re-review

rpignolet (approved):

tdp/cli/conftest.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2022 TOSIT.IO
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from tdp.conftest import generate_collection
+
+
+@pytest.fixture(scope="function")
```
PaulFarault, line 9: `scope="function"` is default with Pytest. You can remove it.

tdp/cli/commands/test_init.py
```diff
@@ -0,0 +1,37 @@
...
+        "--vars",
+        vars,
+    ]
+    runner = CliRunner()
+    result = runner.invoke(init, args)
+    assert result.exit_code == 0
+
+
+def test_tdp_init_db_is_created(collection_path, vars, tmp_path):
+    dp_path = tmp_path / "sqlite.db"
```
PaulFarault, line 25: What does `dp_path` stands for? Did you meant `db_path`?

PaulFarault (changes requested):
Why aren't the tests involving the `plan` commands in `commands/plan`?

tdp/cli/conftest.py
sergkudinov: ok

tdp/cli/commands/test_init.py
sergkudinov: yes, typo

PaulFarault:
Can't the `conftest` be moved to the `commands` directory?

Is it used elsewhere in the cli? If no we may want to move it too.

sergkudinov:
> Can't the `conftest` be moved to the `commands` directory?
> 
> Is it used elsewhere in the cli? If no we may want to move it too.

So far, it is is used only in commands, moving, later on we can move it back if neccessary.

PaulFarault (approved):

PR #352, opened 2023-07-05
fix: null value in deployment deployment log

### LINKED ISSUE #317: `null` in deployment_log table
In the `deploy_log` table, `targets` and `sources` for deployment of type _DAG_ are set to str `"null"` instead of null value.

jpsavy:

Fixes #317 

#### Additional comments

The null value in the target and source columns of the deployment_log table is changed to JSON.NULL < null > after running the command: tdp plan dag

#### Agreements

### DIFF
--- a/tdp/core/models/deployment_log.py
+++ b/tdp/core/models/deployment_log.py
@@ -61,8 +61,8 @@ class DeploymentLog(Base):
     __tablename__ = "deployment_log"
 
     id = Column(Integer, primary_key=True)
-    sources = Column(JSON)
-    targets = Column(JSON)
+    sources = Column(JSON(none_as_null=True))
+    targets = Column(JSON(none_as_null=True))
     filter_expression = Column(String(length=OPERATION_NAME_MAX_LENGTH * 5))
     filter_type = Column(Enum(FilterTypeEnum))
     start_time = Column(DateTime(timezone=False))


PaulFarault (approved):
It has been triggering me for a while. Nice catch!

rpignolet (changes requested):
For me this is not the right way to correct. You have to change the schema so that in the case of a `None` the "SQL null" is used.
Here is what I found on the internet: https://stackoverflow.com/questions/29309696/filter-postgres-json-column-by-null-value-in-sqlalchemy
If you can look to add the `none_as_null=True` at schema level to test.

jpsavy:
> For me this is not the right way to correct. You have to change the schema so that in the case of a `None` the "SQL null" is used. Here is what I found on the internet: https://stackoverflow.com/questions/29309696/filter-postgres-json-column-by-null-value-in-sqlalchemy If you can look to add the `none_as_null=True` at schema level to test.

According to SQlAlchelmy documentation : 
https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.JSON.params.none_as_null
Using Column(JSON(none_as_null=True)) or Column(JSON(none_as_null=True)) for the sources ad targets columns in the DeploymentLog class doesn't work : always null instead of <null> for a 'None' value.

This is why a use this actual workaround : 
targets=null() if targets is None else targets,

tdp/core/deployment/deployment_plan.py
```diff
@@ -1,6 +1,8 @@
 # Copyright 2022 TOSIT.IO
 # SPDX-License-Identifier: Apache-2.0
 
+from sqlalchemy import null,text
```
rpignolet, line 4: Why adding this import ?

jpsavy: > Why adding this import ?

No more import null and text from sqlalchelemy in deployment_plan.py

rpignolet:
LGTM, if you can squash your commit please. Not need to have 3 commits to change 2 lines.

PaulFarault:
I'll do it for @jpsavy as he won't be available for a few weeks.

PaulFarault (approved):

rpignolet (approved):

PR #367, opened 2023-07-11
fix: override collections items type

PaulFarault:

Fixes None

#### Additional comments

Override `Mapping.items` method to specify the correct type.

#### Agreements


mdrutel (approved):

tdp/core/collections.py
```diff
@@ -46,6 +47,9 @@ def __iter__(self):
     def __len__(self):
         return self._collections.__len__()
 
+    def items(self) -> Iterator[Tuple[str, Collection]]:
```
rpignolet, line 50: Why do you implement this method ? It should already be implemented because `Collections` extends `Mapping` no ?

PaulFarault: To have the correct type inferred.

For example:

```py
for collection_name, collection in collections.items():
```

By default, `collection_name` and `collection` are both of type `Any`. Overriding the `Mapping.items` methods allows to specifies the returned types.

rpignolet: Just for type, seriously ...

PaulFarault: I haven't find a better way to do it 🤷🏻‍♂️

rpignolet: And you will do the same for all mixing method of `Mapping` like `keys`, `values` and `get` ? We lose the value of `Mapping` "abc" to only implement required method.

PaulFarault: Well, this wouldn't be a problem with Python 3.9: https://docs.python.org/3.9/library/typing.html#generics

PaulFarault:
Well, this wouldn't be a problem with Python 3.9: https://docs.python.org/3.9/library/typing.html#generics

Will fix it in #369

PR #376, opened 2023-07-20
refactor: collection methods

PaulFarault:
Remove unused methods from collection class.
Rename operations to playbooks to avoid confusion.

Fixes None

#### Additional comments

Refactor the `Collection` class:

- Remove unused properties (most of the paths are only used internally). Create a `_get_directory` method to handle the retrieval of dir without these properties.
- Rename "operations" to "playbooks" to better describe the objects in this context.
- Remove the `from_path` static method as it don't bring much (it takes the same parameters as `__init__`). The tests have been moved into their own method (it should make unit testing easier).

#### Agreements

### DIFF
--- a/tdp/core/collection.py
+++ b/tdp/core/collection.py
@@ -4,11 +4,11 @@
 import json
 import os
 from pathlib import Path
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union, Tuple
 
 DAG_DIRECTORY_NAME = "tdp_lib_dag"
 DEFAULT_VARS_DIRECTORY_NAME = "tdp_vars_defaults"
-OPERATION_DIRECTORY_NAME = "playbooks"
+PLAYBOOKS_DIRECTORY_NAME = "playbooks"
 SCHEMA_VARS_DIRECTORY_NAME = "tdp_vars_schema"
 
 JSON_EXTENSION = ".json"
@@ -17,7 +17,7 @@ YML_EXTENSION = ".yml"
 MANDATORY_DIRECTORIES = [
     DAG_DIRECTORY_NAME,
     DEFAULT_VARS_DIRECTORY_NAME,
-    OPERATION_DIRECTORY_NAME,
+    PLAYBOOKS_DIRECTORY_NAME,
 ]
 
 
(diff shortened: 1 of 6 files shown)


tdp/core/collection.py
```diff
@@ -37,128 +37,74 @@ class Collection:
...
-        return self._path
+        self._services: Optional[List[Path]] = None
 
     @property
     def name(self) -> str:
         """Name of the collection."""
         return self._path.name
 
-    @property
-    def dag_directory(self) -> Path:
```
rpignolet, line 101: Why removing theses methods ? Just because it's not used outside the class doesn't mean it should be deleted.

tdp/core/collection.py
```diff
@@ -37,128 +37,74 @@ class Collection:
...
             self._operations = {
                 playbook.stem: playbook
-                for playbook in self.operations_directory.glob("*" + YML_EXTENSION)
+                for playbook in self._get_directory(PLAYBOOKS_DIRECTORY_NAME).glob(
+                    "*" + YML_EXTENSION
+                )
             }
         return self._operations
 
-    def get_service_default_vars(self, service_name: str) -> List[Path]:
```
rpignolet, line 137: This is used by `tdp default-diff` command.

tdp/core/collection.py
```diff
@@ -37,128 +37,74 @@ class Collection:
...
-        """
-        service_path = self.default_vars_directory / service_name
-        if not service_path.exists():
-            return []
-        return [(path.name, path) for path in service_path.glob("*" + YML_EXTENSION)]
-
     def get_service_schema(self, service_name: str) -> Dict:
-        """Get the schema of the service.
-
-        Args:
```
rpignolet, line 154: Why removing `Args` and `Returns` doc ?

rpignolet (changes requested):

tdp/core/collection.py
PaulFarault: They don't provide much value, but it'll better to fix them later along with doc refactoring.

tdp/core/collection.py
PaulFarault: True, idk why my linter didn't warned me / test were still passing.

tdp/core/collection.py
```diff
@@ -36,67 +36,47 @@ class MissingMandatoryDirectoryError(Exception):
...
-        if not path.is_dir():
-            raise PathIsNotADirectoryError(f"{path} is not a directory")
-        for mandatory_directory in MANDATORY_DIRECTORIES:
-            mandatory_path = path / mandatory_directory
-            if not mandatory_path.exists() or not mandatory_path.is_dir():
-                raise MissingMandatoryDirectoryError(
-                    f"{path} does not contain the mandatory directory {mandatory_directory}",
-                )
-        return Collection(path)
+        return Collection.from_path(path=Path(path).expanduser().resolve())
```
rpignolet, line 68: I don't know how it works but for me this is an infinite loop. Have you tested your code ?

rpignolet (approved):

PR #389, opened 2023-08-09
feat: add extra vars feature for operation/playbook

### LINKED ISSUE #195: Enable use of playbook parameters at runtime
This issue is to discuss enabling parameters to be added when deploying tasks (like with [--extra-vars](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#defining-variables-at-runtime) in playbooks deployment). 
This would enable different deployments based on a same role / playbooks, as well as allowing for a more dynamic use of roles. 

- For instance, one could dynamically create user defined with the parameter "username" on top of the role `ansible-hadoop-users`:
    
    ```bash
    tdp deploy users-install -e "user=gonzalo"
    ```
    
 - Or have a single playbook use different set of vars based on the argument provided and solving the used vars with a [lookup](https://serverfault.com/questions/907164/ansible-conditionally-define-variables-in-vars-file-if-a-certain-condition-is-m). For instance, define different hosts architectures:
    
    ```bash
    tdp deploy generate-node-deployment -e "architecture=getting_started"
    ```
    
    and 
    
    ```yml
    hosts: "{{ lookup('vars','hosts_' + architecture) }}"
    ```

I understand this might not be the scope of tdp-lib as complete DAGs have to be implemented and might complicate dependencies. But the implementation still has some interesting uses.

rpignolet:

Fixes #195 

#### Agreements

### DIFF
--- a/tdp/cli/commands/plan/run.py
+++ b/tdp/cli/commands/plan/run.py
@@ -18,12 +18,21 @@ from tdp.core.models import DeploymentLog
     multiple=True,
     help="Hosts where operations are launched. Can be used multiple times.",
 )
+@click.option(
+    "-e",
+    "--extra-vars",
+    envvar="TDP_EXTRA_VARS",
+    type=str,
+    multiple=True,
+    help="Extra vars for operations (forwarded to ansible as is). Can be used multiple times.",
+)
 @collections
 @database_dsn
 @vars
 def run(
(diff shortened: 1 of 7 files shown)


PaulFarault (approved):

mdrutel (approved):
Tested OK

PR #403, opened 2023-08-17
refactor: dissociate operation name from operation

PaulFarault:

Fixes None

#### Additional comments

- Extract `OperationName` from `Operation` to ensure that every class answer to only one objective. Previously, the `Operation` class was used for both parsing operation name and instanciate operations to run. `OperationName` class inherit from the current `ServiceComponentName` class to add an `action_name` to it. The `Operation` class now uses an `OperationName` instance as a name.
- Make all attributes of the `Operation` class private and create `@property` to access them (once instanciated, operations should be immutables).
- Create specific errors (`InvalidOperationNameError`, `InvalidServiceNameError`, `InvalidComponentNameError`, `InvalidActionNameError` and `InvalidHostNameError`).
- Create test to cover all use cases of the 3 classes (`ServiceComponentName`, `OperationName`, `Operation`).
- Add types and documentation.

#### Agreements


sergkudinov (approved):
Don't see anything wierd. Tests are passing.

tdp/core/operation.py
mdrutel, line 1: I don't know if it's the right place to discuss it, but if I wanted to be strict, i would say that OPERATION_NAME_MAX_LENGTH would need 1 or 2 more characters to take into account the "_" chars ^^'

PaulFarault: You're right, we should add `+ 2` to the `OPERATION_NAME_MAX_LENGTH`. I'll do it if we agree on this refactor!

tdp/core/operation.py
```diff
@@ -23,97 +16,279 @@
...
+                f"Operation name '{self.full_name}' must be less than "
+                + "{OPERATION_NAME_MAX_LENGTH} characters."
+            )
+
+    @property
+    def full_name(self) -> str:
+        """Full name of the operation."""
+        return f"{super().full_name}_{self.action_name}"
+
+    def is_service_opearion(self) -> bool:
```
mdrutel, line 159: Though it is not used, I think we have to correct the method name ^^'

tdp/core/operation.py
```diff
@@ -23,97 +16,279 @@
...
-        action: Name of the action.
-        name: Name of the operation.
-        collection_name: Name of the collection where the operation is defined.
-        component: Name of the component.
-        depends_on: List of operations that must be executed before this one.
-        noop: If True, the operation will not be executed.
-        service: Name of the service.
-        host_names: Set of host names where the operation can be launched.
+    An operation can be defined in an Ansible collection. It can either be part of the
+    DAG or not (if it has some depencies defined in the `tdp_lib_dag` directory).
```
mdrutel, line 207: I think it's "dependencies" (sry ^^')

PR #412, opened 2023-08-23
Test db

PaulFarault:

Fixes None

#### Additional comments

- Refactor tests to enable query testing.
- Add tests for the `get_latest_success_component_version_log` query.

#### Agreements


tdp/core/models/test_models.py
```diff
@@ -0,0 +1,82 @@
...
+        state="Success",
+        logs=b"operation log",
+    )
+
+    deployment_log.component_version.append(component_version_log)
+    deployment_log.operations.append(operation_log)
+
+    logger.info(deployment_log)
+    logger.info(operation_log)
+    logger.info(operation_log)
```
sergkudinov, line 52: You may mean `logger.info(component_version_log)` the second time?

sergkudinov (changes requested):

tdp/core/models/test_models.py
PaulFarault: Didn't spot it as it was already there but you're right. I fixed it!

sergkudinov (approved):

tdp/conftest.py
```diff
@@ -1,17 +1,48 @@
...
     DAG_DIRECTORY_NAME,
     DEFAULT_VARS_DIRECTORY_NAME,
     PLAYBOOKS_DIRECTORY_NAME,
     SCHEMA_VARS_DIRECTORY_NAME,
 )
+from tdp.core.models import Base
+
+
+sqlalchemy_logger = logging.getLogger("sqlalchemy.engine")
+sqlalchemy_logger.setLevel(logging.INFO)
```
rpignolet, line 23: Where this logger is used ?

PaulFarault: It is not used in my code, it is a way to define log level as "INFO" for sqlalchemy in those tests.

We could replace it with:

```py
logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
```

rpignolet: ok so no need to put it in a variable.

PaulFarault: Beyond the syntax, idk if we should keep it.

PaulFarault: I removed it.

rpignolet (approved):

PR #420, opened 2023-08-31
refactor: replace OrderedDict with dict

### LINKED ISSUE #369: Fix types when migrating to Python 3.9
Upgrading to Python 3.9 allows us to fix a lot of types. In particular:

- Remove `OrderedDict`
- Explicit the correct types for classes that inherit from the `Mapping` abc (as `Collections`)
- Use `list` instead of `List`
- etc

sergkudinov:

Part of #369

#### Additional comments

This seams to be already fixed:

> - Explicit the correct types for classes that inherit from the `Mapping` abc (as `Collections`)
> - Use `list` instead of `List`

I am not sure if there is smth else to refactor. Checked with [whats new in python 3.9](https://docs.python.org/3/whatsnew/3.9.html).

#### Agreements

### DIFF
--- a/tdp/core/variables/service_variables.py
+++ b/tdp/core/variables/service_variables.py
@@ -7,7 +7,7 @@ import os
 from collections import OrderedDict
 from contextlib import ExitStack, contextmanager
 from pathlib import Path
-from typing import TYPE_CHECKING, Iterator, Union
+from typing import TYPE_CHECKING, Generator, Union
 
 import jsonschema
 from jsonschema import exceptions
@@ -180,7 +180,7 @@ class ServiceVariables:
     @contextmanager
     def open_var_files(
         self, message: str, paths: list[str], fail_if_does_not_exist: bool = False
-    ) -> Iterator[OrderedDict[str, Variables]]:
+    ) -> Generator[OrderedDict[str, Variables], None, None]:
         """Open variables files.
 
         Adds the underlying files for validation.
--- a/tdp/core/variables/service_variables.py
+++ b/tdp/core/variables/service_variables.py
@@ -180,9 +180,7 @@ class ServiceVariables:
     @contextmanager
     def open_var_files(
         self, message: str, paths: list[str], fail_if_does_not_exist: bool = False
-    ) -> Iterator[
-        dict[str, Variables]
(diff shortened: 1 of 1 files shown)


rpignolet:
I don't know why you want to use `dict` instead of `OrderedDict`.

sergkudinov:
In the issue you mention: 

> Remove `OrderedDict`

I understood we replace `OrderedDict`, which `dict` does the same since Python 3.6, it also keeps items ordered ([reference](https://realpython.com/python-ordereddict/#choosing-between-ordereddict-and-dict)) and a bit more efficiente than `OrderedDict`.

To my mind, it worth to do that.

sergkudinov:
@PaulFarault @rpignolet 
what do you think?

rpignolet:
I think is better to keep `OrderedDict` to no rely on detail implementation. Moreover, with `OrderedDict` it's explicit that we want an order for keys.

PaulFarault:
Thx. As we're looking at it, could you add a commit to fix the Iterator please (make it a Generator).

rpignolet (approved):

tdp/core/variables/service_variables.py
```diff
@@ -180,9 +180,7 @@ def _open_var_file(
     @contextmanager
     def open_var_files(
         self, message: str, paths: list[str], fail_if_does_not_exist: bool = False
-    ) -> Iterator[
-        dict[str, Variables]
-    ]:  # TODO: Transform Dict to OrderedDict with python>3.6
+    ) -> Generator[OrderedDict[str, Variables]]:
```
PaulFarault, line 183: Generator generic takes 3 type parameters: `Generator[YieldType, SendType, ReturnType]`.

```suggestion
    ) -> Generator[OrderedDict[str, Variables], None, None]:
```

PaulFarault (changes requested):
I saw that you linked this PR to #369 which will close it. However, it seems that there are still some types to migrate. Could you please either fix the whole issue or transform the `Fixes` keyword to `Part of`?

tdp/core/variables/service_variables.py
sergkudinov: done

PaulFarault (approved):

PR #287, opened 2022-12-12
Cli minor refactors

gboutry:
#### Additional comments

This PR brings minor refactors such as:

- Centralize common tdp options definition
- make `.env` configurable via an option
- move query outside of commands package

#### Agreements

- [X] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0).
- [X] I hereby agree to grant [TOSIT](https://www.tosit.io/) a copyright license to use my contributions.

PR #425, opened 2023-09-01
423 `tdp` command with no arguments should print the usage help

### LINKED ISSUE #423: `tdp` command with no arguments should print the usage help
just a very small nitpick. When using `tdp` command with no argument, the command just returns. It should print the help same as `tdp -h`

PACordonnier:

Fixes #423

#### Additional comments

This fixes the issue by using the default click behaviour being 

> By default, a group or multi command is not invoked unless a subcommand is passed. In fact, not providing a command automatically passes --help by default

If we do not want such behaviour, we should implement a context such as this:

```
@click.pass_context
def cli(ctx):
    if ctx.invoked_subcommand is None:
        click.echo('I was invoked without subcommand')
    else:
        click.echo('I am about to invoke %s' % ctx.invoked_subcommand)
```

It is a quick fix for a very minor behaviour. I haven't seen in the history why this behaviour was added (in PR #287)

Link: https://click.palletsprojects.com/en/7.x/commands/

#### Agreements

### DIFF
--- a/tdp/cli/__main__.py
+++ b/tdp/cli/__main__.py
@@ -21,7 +21,7 @@ from tdp.cli.commands.validate import validate
 CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 
 
-@click.group(context_settings=CONTEXT_SETTINGS, invoke_without_command=True)
+@click.group(context_settings=CONTEXT_SETTINGS)
 @click.option(
     "--env",
     default=".env",


sergkudinov (approved):
I think printing just help message is fine.

gonzaloetjo (approved):

PaulFarault (approved):

rpignolet (approved):

PR #461, opened 2023-10-11
Alow to override stale status

### LINKED ISSUE #348: [meta] List stale components configurations per host to split a reconfigure
[We want to be able to split the generation of the deployment plan into several times (#308).

To achieve this, we need to track stales configurations. Hence, to persist the state of each component/host that have been edited in the database. The following tables are required:

- **stale_component**: to track the states of the stale component in the context of a reconfigure.

  ```txt
  service_name | component_name | to_reconfigure | to_restart | (to_check, to add later) | (node, to add later)
  ```

- **component_versions_log**: the actual service_component_log table, to link the configurations version to a deployment.

  ```txt
  deployment_id | service_name | component_name | config_version | (node, to add later)
  ```

Adding a `stale` command is required to interact with the new table, with the following sub-commands:

- `tdp stale`, print a list of operations to reconfigure the components from the stale_component table.
- `tdp stale --generate`, to populate the stale_component table.
- `tdp stale add`, to add a service or component to the stale list.
- `tdp stale remove` to remove a service or component from the stale list (for advanced users).

Ideally, the stale_component table should be populated at each configuration edition. This requires a specific command to edit the configurations. However, the `tdp stale --generate` command could be used in the event where a config is edited by hand.

Then, the `tdp plan reconfigure` command should be re-implemented so that it can use the stale_component table.

---

#### Status

- [x] Refactor the service_component_log to the component_version_log table (#353)
- [X] Refactor the `ServiceVariables.component_modified` (#362 )
- [x] Create the `tdp stale` command with the `--generate` option and the stale_component table (#378)
- [ ] Allow to override stale status (#461)
- [x] Rewrite the `tdp plan reconfigure` command to use the stale_component table (#379)

### LINKED ISSUE #465: Stale status edition message
Once #461 is merged.

User should be able to enter a message to describe their changes when manually editing the stale status of a component.

### LINKED ISSUE #470: Exclude cluster manipulation commands from `tdp status`
We are used to using the 'status' command in any CLI just for displaying the actual state or other information, but not to manipulate with entities. For example, `git status` or `docker info` (`info` has the same meaning).

Currently, with `tdp status —generate-stales` and `—override` we operate the cluster . It's misleading having such subcommands/options in `status`.

The current usage of `tdp status`:

```
Usage: tdp status [OPTIONS] [SERVICE] [COMPONENT]

Options:
  --collection-path TEXT      List of paths separated by your os' path
                              separator (:)  [required]
  --database-dsn TEXT         Database Data Source Name, in sqlalchemy driver
                              form example: sqlite:////data/tdp.db or
                              sqlite+pysqlite:////data/tdp.db. You might need
                              to install the relevant driver to your
                              installation (such as psycopg2 for postgresql)
                              [required]
  --generate-stales           Update the list of stale components.
  --host TEXT                 Hosts where components are defined. Can be used
                              multiple times.
  -O, --override              Override the stale status.
  --stale                     Only print stale components.
  --to-config BOOLEAN         To be used with --override. Override to_config
                              value.
  --to-restart BOOLEAN        To be used with --override. Override to_restart
                              value.
  --validate / --no-validate  Should the command validate service variables
                              against defined JSON schemas
  --vars PATH                 Path to the tdp vars  [required]
  -h, --help                  Show this message and exit.
```

## Suggestion

Move the `--generate-stales`, `--override`, `--to-config`, `--to-restart` options to another command, for example `tdp stale` (as it was before). Also, the `--generate-stales` option is redundant in this case, just `tdp stale` can take over the `tdp stale --generate-stales` functionality.

PaulFarault:

Fixes #348 
Fixes #465 
Fixes #470

#### Additional comments

Allow to manually edit `to_config` and `to_restart` status properties.

#### Agreements


tdp/cli/commands/status/generate_stales.py
```diff
@@ -0,0 +1,59 @@
...
+    service: Optional[str],
+    component: Optional[str],
+    collections: Collections,
+    database_dsn: str,
+    validate: bool,
+    vars: Path,
+) -> None:
+    """Generate stale components.
+
+    Stales components are components that have been modified and need to be
```
rpignolet, line 38: For me it is not clear the definition of a modified component. It is necessary to write that a component is modified when its associated variables are modified.

tdp/cli/utils.py
```diff
@@ -79,6 +80,7 @@ def collections(func: FC) -> FC:
         required=True,
         callback=_collections_from_paths,
         help=f"List of paths separated by your os' path separator ({os.pathsep})",
+        is_eager=True,
```
rpignolet, line 83: Why adding `eager` ?

PaulFarault: So that this option is loaded first. It can then be consumed by others, as this one for example:

```py
def _check_component(
    ctx: click.Context, param: click.Parameter, value: Optional[str]
) -> Optional[str]:
    """Click callback that check if the component exists."""
    collections: Collections = ctx.params["collections"]
    service: str = ctx.params["service"]
    if value and value not in [
        sc_name.component_name
        for sc_name in collections.get_components_from_service(service)
    ]:
        raise click.UsageError(
            f"Component '{value}' does not exists in service '{service}'."
        )
    return value
```

I only added `is_eager=True` to the ones I needed. But, I think that we should add it on any "primitive" option that we consider (basically, the one we put in the .env file in my opinion).

tdp/cli/commands/status/generate_stales.py
PaulFarault: I suggest we improve the documentation after having refactored every commands. My argument is that other commands use the notion of stale, some core methods uses this notion too, and I'd rather write something coherent for every methods at once.

Do you agree with that or do you want me to improve this one now?

rpignolet: ok

tdp/cli/utils.py
rpignolet: Add a comment to know why `eager` is added.

PaulFarault: Done!

rpignolet (approved):

mdrutel (approved):
tested OK.
Maybe we could display messages written by the user with the option ```-m```

PaulFarault:
> Maybe we could display messages written by the user with the option -m

I think that we should display both:

- The last edition date in the status table
- The history of editions, using a `--history` option

PR #481, opened 2023-10-26
issue 346 transformed playbook.py into script instead of TDP cli

### LINKED ISSUE #346: [cli] Move the `playbooks` command to a script
The `tdp playbooks` command is used to generate meta playbooks, which are employed for deploying TDP without the necessity of using tdp-lib. We could convert it to a script instead of having it as a command.

The underlying question may be: "what's the purpose of tdp-lib?". Is it made for developers or for users?

SteBaum:

Fixes #346

#### Additional comments

- Removed the playbook command in main.py.
- Transformed playbook.py into a script a developer can use outside of tdp lib commands.
- Changed `import_playbook` to  `ansible.builtin.import_playbook` for output yml files.

#### Agreements


scripts/playbooks.py
```diff
@@ -121,3 +114,44 @@ def write_copyright_licence_headers(fd):
...
+        help="Only write operation from this collection, specify the collection name",
+    )
+    args = parser.parse_args()
+    services = args.services
+    output_dir = args.output_dir
+    for_collection = args.for_collection
+    collections = Collections.from_collection_list(
+        [
+            Collection.from_path(col)
+            for col in args.collection or TDP_COLLECTION_PATH.split(":")
```
PaulFarault, line 148: You should use `os.pathsep` as a separator.

scripts/playbooks.py
```diff
@@ -1,37 +1,28 @@
...
-    type=str,
-    help="Only write operation from this collection",
-    required=False,
-    multiple=True,
-)
-@collections
-def playbooks(services, output_dir, for_collection, collections):
+
+def playbooks(
+    collections: Collections,
```
PaulFarault, line 20: Add a condition before calling playbooks if you don't want `collections` to be empty.

scripts/playbooks.py
PaulFarault, line 1: Would be nice to add a docstring describing the purpose of this script

PaulFarault (changes requested):

PaulFarault:
We could leverage click's features to parse options (and use the ones defined for the cli). This doesn't mean that the command will is part of the CLI.

PaulFarault (dismissed):
We could leverage click's features to parse options (and use the ones defined for the cli). This doesn't mean that the command will is part of the CLI.

PaulFarault (changes requested):

SteBaum:
I have changed the PR. It now does what is described in the description above and does not change failure if the meta outputdirectory already exists.

rpignolet (approved):

PR #484, opened 2023-10-26
feat: prevent default vars copy

### LINKED ISSUE #320: Do not copy tdp_vars_default in tdp_vars
While doing the `tdp init` command, the `tdp_vars_default` are copied inside `tdp_vars`. It would be better to keep them in a distinct folder to facilitate upgrades and reset.

PaulFarault:

Fixes #320

#### Agreements

### DIFF
--- a/tdp/core/collections.py
+++ b/tdp/core/collections.py
@@ -197,16 +197,6 @@ class Collections(Mapping[str, Collection]):
                     collection_name=collection_name,
                 )
 
-    def get_services(self) -> set[str]:
-        """Get the list of services names."""
-        # Filters out the sleep operation service as it is not a real service.
-        operation_sleep_name = ServiceComponentName(OPERATION_SLEEP_NAME)
-        return {
-            operation.service_name
-            for operation in self.operations.values()
-            if operation.service_name != operation_sleep_name.service_name
-        }
-
     def get_default_vars_paths(self, service_name: str) -> Generator[Path, None, None]:
         """Get the default vars paths for a service."""
         generators = (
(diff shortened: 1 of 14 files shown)


PaulFarault:
This is a work in progress. I still need to merge tdp_vars with all tdp_vars_default to be able to validate the variables' schema.

tdp/core/collections.py
```diff
@@ -195,6 +196,19 @@ def _init_operations(self):
...
+        return {operation.service_name for operation in self.operations.values()}
+
+    def get_default_vars_path(
+        self, service_name: str
+    ) -> Optional[Generator[Path, None, None]]:
+        """Get the default vars paths for a service."""
+        for collection in self._collections.values():
+            default_vars = collection.get_service_default_vars(service_name)
+            if default_vars:
+                return default_vars
```
rpignolet, line 210: Maybe you want to `yield` here instead of `return` ?

PaulFarault: `Collection.get_service_default_vars` returns a Generator.

Replacing `return` by `yield` would lead to a `Generator[Generator[Path, None, None], None, None]`, I'm not sure that what we want to obtain.

rpignolet: Indeed but you stop the loop over collections at the first collection where there are defaults vars, a service can have default vars in multiple collection.

PaulFarault: I fixed it!

PaulFarault:
We should base the repository creation on the default variables instead of the collections. This ensure that the service is expecting some variables (as services can be artificially added by users, without requiring any vars).

PR #491, opened 2023-10-30
Check operation availability on host before execution

### LINKED ISSUE #490: Check that operation is available for an host before execution
If a host is specified for an operation, we should check if the operation is available on this host before launching the Ansible command (which uses the `--limit` option).

PaulFarault:

Fixes #490 

#### Additional comments

If a host is provided for an operation, check if the operation is available for this host before executing it.
Log a custom error message in the database.

This is required to avoid edge case with the `--limit <host>` option on the `ansible-playbook` command.

#### Agreements

### DIFF
--- a/tdp/core/deployment/deployment_runner.py
+++ b/tdp/core/deployment/deployment_runner.py
@@ -52,6 +52,20 @@ class DeploymentRunner:
         operation_log.start_time = datetime.utcnow()
 
         operation = self._collections.get_operation(operation_log.operation)
+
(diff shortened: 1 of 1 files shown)


sergkudinov:
Would be good to add a unit test for this case.

sergkudinov:
How to reproduce/to test manually?

rpignolet (approved):

PaulFarault:
> How to reproduce/to test manually?

Try to run an operation on an host that is not defined for this operation.

Issue #340, opened 2023-06-30
Database left in an incorect state if interrupting an ongoing deployment


PaulFarault:
An interrupted deployment keeps its status as `RUNNING`.

Is it a normal behavior? How to deal with this case as nothing failed but a few operations already succeeded?

PaulFarault:
It seems normal to left the database "as is" in this case. What we should look for is how to recover from this state (maybe using a `resume`?).

sergkudinov:
To my mind, if we have `resume`, it is naturally to have a `pause` command. In the context of TDP, it would be very useful.

Deployment process could be interrupted by the user in 2 manners:
- sending a temination signal with `CTRL+C` or clossing the parrent process
- running a command, for example `tdp pause` command or via tdp-server in the future.

Having `pause`, we will need another status `PAUSED` for deployments and operations. That would complement this current issue with a better usage.

PR #492, opened 2023-10-31
Resolve `ansible-playbook` command before execution

PaulFarault:

Fixes None

#### Additional comments

- use += operator to compose the command
- resolve `ansible-playbook` command before executing the script

#### Agreements

### DIFF
--- a/tdp/core/deployment/executor.py
+++ b/tdp/core/deployment/executor.py
@@ -3,6 +3,7 @@
 
 import io
 import logging
+import shutil
 import subprocess
 from collections.abc import Iterable
 from typing import Optional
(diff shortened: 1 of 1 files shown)


sergkudinov (approved):

sergkudinov:
Behavior of `tdp deploy` is wrong when it is interrupted by the error of missing ansible.

The problem is that the status of the current deployment is changed to `RUNNING`, but the issue caused this is of a "global" scope, because we shouldn't run the deployment until we have checked all the prerequisites.

To reproduce, first simulate the absence of `ansible-playbook` in PATH, and then run:

```
tdp plan dag
tdp deploy --mock-deploy
tdp browse
```

I am not sure whether it is in the scope of this PR, probably not.

PaulFarault:
In my understanding you are mentioning 2 différents issues:

- The status that stays `RUNNING` after a failure, which will be handled in #340 
- Requirements testing, which we should discuss at some point (no existing issue)

PR #502, opened 2023-11-09
Add color on ansible logs

PaulFarault:

Fixes None

#### Additional comments

Add `ANSIBLE_FORCE_COLOR` environment variable to the ansible subprocess to enable color on logs.

#### Agreements

### DIFF
--- a/tdp/core/deployment/executor.py
+++ b/tdp/core/deployment/executor.py
@@ -3,6 +3,7 @@
 
 import io
 import logging
+import os
 import shutil
 import subprocess
 from collections.abc import Iterable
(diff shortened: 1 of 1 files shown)


rpignolet:
Why forcing Ansible color ? If a user wants color, it can be activate by adding `ANSIBLE_FORCE_COLOR` inside `.env` no ?

PaulFarault:
I'd like to find a way to both:

- be able to remove the colors from the output (as it is currently by default)
- but still being able to enable them later when browsing the logs

If we omit them by default, we would never be able to retrieve them later.

PaulFarault:
We should use the `ansible.cfg` config, no need to force it.

PR #505, opened 2023-11-13
Prevent failure when rolling interval on `tdp plan ops`

PaulFarault:

Fixes None

#### Additional comments

The `tdp plan ops` command used with the `--rolling-interval` option caused the deployment to failed because of an host issue. Host should not be specified for the `wait_sleep` playbook.

#### Agreements

### DIFF
--- a/tdp/core/models/deployment_model.py
+++ b/tdp/core/models/deployment_model.py
@@ -240,7 +240,7 @@ class DeploymentModel(BaseModel):
                         OperationModel(
                             operation=OPERATION_SLEEP_NAME,
                             operation_order=i,
-                            host=host_name,
+                            host=None,
                             extra_vars=[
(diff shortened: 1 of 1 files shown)


tdp/core/models/deployment_model.py
```diff
@@ -240,7 +240,6 @@ def from_operations(
                         OperationModel(
                             operation=OPERATION_SLEEP_NAME,
                             operation_order=i,
-                            host=host_name,
```
rpignolet, line 243: `host` should be set to `None`  because it will change `host` value if the Operation is updated in the database.

PaulFarault: You're right.

rpignolet (approved):

PR #522, opened 2023-11-22
Check if file is writable before flushing

PaulFarault:

Fixes None

#### Additional comments

IMO it makes more sense to check if the file is writable before flushing on disk.

Flushing on disk on a non writable file should raise an error.

#### Agreements

### DIFF
--- a/tdp/core/variables/variables.py
+++ b/tdp/core/variables/variables.py
@@ -118,7 +118,7 @@ class _VariablesIOWrapper(VariablesDict):
         return proxy(self)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self._close()
+        self.close()
 
     def _flush_on_disk(self) -> None:
         """Write the content of the variables file on disk.
@@ -146,7 +146,7 @@ class _VariablesIOWrapper(VariablesDict):
         # https://docs.python.org/3/library/os.html#os.fsync
         os.fsync(self._file_descriptor.fileno())
 
-    def _close(self) -> None:
+    def close(self) -> None:
         """Closes the file descriptor.
 
         Raises:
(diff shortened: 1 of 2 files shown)


tdp/core/variables/variables.py
```diff
@@ -118,22 +118,25 @@ def __enter__(self):
...
             RuntimeError: If the file descriptor is already closed.
         """
+        # Check if the file descriptor is already closed
         if not self._file_descriptor or self._file_descriptor.closed:
             raise RuntimeError(
                 f"{self._file_path} is already closed, which shouldn't happen"
             )
 
+        # Check if the file is writable
         if not self._file_descriptor.writable():
```
sergkudinov, line 136: If `_flush_on_disk` is only used in `_close`, this condition is never passing, because this is already checked [here](https://github.com/TOSIT-IO/tdp-lib/pull/522/files#diff-60263f9c86f45f8ceaaabbdb576ec29dd446d11c610fa7d7e114bafd5f925280R162)

PaulFarault: Agree, I left it to prevent a bad usage of  `_flush_on_disk`.

sergkudinov:
I added a unit test for that

sergkudinov (approved):

SteBaum (approved):

tdp/core/variables/variables.py
```diff
@@ -143,17 +146,22 @@ def _flush_on_disk(self) -> None:
         # https://docs.python.org/3/library/os.html#os.fsync
         os.fsync(self._file_descriptor.fileno())
 
-    def close(self) -> None:
+    def _close(self) -> None:
```
rpignolet, line 149: This is not a private method, if a user do not want to use `with`, he must use this `close` method like the `open` function in Python.

PaulFarault: Fixed!

rpignolet (approved):

PR #523, opened 2023-11-24
feat: Define log level based on a config file

### LINKED ISSUE #516: Define log level based on a config file
For now, log level is set globally through the `--log-level` option on the CLI. The whole program and its dependency use this level.

It will be better to configure it [using a file](https://docs.python.org/3/library/logging.config.html#module-logging.config), to have more control over log level for each module (GitPython is very verbose in debug mode), and to benefit from it when using only the core lib.

mdrutel:

Fixes #516 

#### Additional comments

With this PR, it will be up to the user to define precisely what he wants for the logging.
The default behaviour is the old one.

#### Agreements


PaulFarault:
Close as stale

PR #527, opened 2023-11-24
`ServiceVariables`: do not store service schema

PaulFarault:

Fixes None

There is no need to store the schema in the `ServiceVariables` class. It can be retrieve from the collections object and passed as an argument when needed.

#### Agreements

### DIFF
--- a/tdp/core/variables/service_variables.py
+++ b/tdp/core/variables/service_variables.py
@@ -10,7 +10,7 @@ from contextlib import ExitStack, contextmanager
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from tdp.core.constants import SERVICE_NAME_MAX_LENGTH, YML_EXTENSION
+from tdp.core.constants import YML_EXTENSION
 from tdp.core.types import PathLike
 from tdp.core.variables.schema import validate_against_schema
 from tdp.core.variables.variables import (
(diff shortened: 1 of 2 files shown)


SteBaum (approved):

mdrutel (approved):
tested OK

tdp/core/variables/service_variables.py
```diff
@@ -242,7 +236,7 @@ def validate_schema(self, variables: Variables, schema: dict) -> None:
         except exceptions.ValidationError as e:
             raise InvalidSchema("Schema is invalid", variables.name) from e
 
-    def validate(self) -> None:
+    def validate(self, schema: dict) -> None:
```
rpignolet, line 239: I don't understand this PR. Now when I have an instance of `ServiceVariables` I need to give a `dict` representing the schema every time I want to validate the variables? But the schema doesn't change between each validation.

PaulFarault: With #320 in mind (preventing default variables copy), we'll need to have access to default variables in the schema validation logic. Both default variables and schema come from the `Collections` object, hence, it seemed better for me to pass a `collections` object as an argument instead of storing everything in the `ServiceVariables` object.

I did this PR to prepare this change.

PR #533, opened 2023-12-07
Disable variable validation by default

PaulFarault:

Fixes None

Disable variable validation against a schema as it is not stable for now.

#### Agreements

### DIFF
--- a/tdp/cli/utils.py
+++ b/tdp/cli/utils.py
@@ -180,7 +180,8 @@ def rolling_interval(func: FC) -> FC:
 def validate(func: FC) -> FC:
     return click.option(
         "--validate/--no-validate",
-        default=True,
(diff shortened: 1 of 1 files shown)


tdp/cli/utils.py
```diff
@@ -180,7 +180,6 @@ def rolling_interval(func: FC) -> FC:
 def validate(func: FC) -> FC:
     return click.option(
         "--validate/--no-validate",
-        default=True,
```
rpignolet, line 183: Put this line in comment and add a TODO ? We will revert it to `True` when the variables validation will work.

PaulFarault: Yep you're right, that's better.

rpignolet (approved):

PR #536, opened 2024-01-16
Refactor Variables class

PaulFarault:

Fixes None

#### Additional comments

Remove the `create_if_missing` option from the `Variables` class initializer as file creation should be handled by the opening mode.

#### Agreements


PaulFarault:
If we want to keep classes, we could also rename the `_VariablesIOWrapper` class to `VariablesFile` and move the `open` method inside.

PR #539, opened 2024-01-17
feat: added query returning the running deployment and the operation record by operation order

### LINKED ISSUE #538: Create a query showing a RUNNING deployment
The cli does not have any query showing a deployment in state `RUNNING` yet. This functionality will be used for an endpoint in the server, so that the user can see if there is a running deployment going on. However, it should be incorporated with the other queries in the cli and the server imports it when it needs it.

SteBaum:

Fixes #538 

#### Additional comments

- Added a query that returns the running deployment if it exists. This query will be used for an endpoint of TDP server.
- Added a query that returns the matching operation record by operation order and deployment id.

#### Agreements

### DIFF
--- a/tdp/cli/queries.py
+++ b/tdp/cli/queries.py
@@ -447,6 +447,30 @@ def get_planned_deployment(session: Session) -> Optional[DeploymentModel]:
     return session.query(DeploymentModel).filter_by(state="PLANNED").one_or_none()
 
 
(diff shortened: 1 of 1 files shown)


PaulFarault:
We should put this PR on hold for 2 reasons:

- `tdp-server` is not ready yet to implement the features requiring this PR
- we need to extract the queries from the cli

PR #542, opened 2024-01-23
feat: added a raise Exception message No planned deployment if none

SteBaum:

Fixes None

#### Additional comments

Added a raise Exception message `No planned deployment` if there is none.

#### Agreements


PaulFarault (commented):
The `try`/`except` doesn't provide any value as it is only switching a `NoResultFound` exception to a very generic `Exception` type.

PR #544, opened 2024-01-29
refactor: changed __next__ iterator returning a callable

### LINKED ISSUE #517: Missing `RUNNING` status for operations
First commit has no effect in the following code. This is expected as the `DeploymentIterator` only return the result of the operation execution.

https://github.com/TOSIT-IO/tdp-lib/blob/abffc4029e749d7bdbb1c9b078134a92a6bffee8/tdp/cli/commands/deploy.py#L99-L103

It would be better if the `DeploymentIterator` was returning an executable, so that we can commit the status in the database before and after execution.

It could even let us return a stream of logs that could be streamed by the server.

SteBaum:

Fixes #517 

#### Additional comments

- Changed the `__next__` function of the deployment_iterator to return a callable.
- The current operation running is marked as `RUNNING` in the operation's database.

#### Agreements

### DIFF
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -90,14 +90,15 @@ def deploy(
 
         if dry:
             for _ in deployment_iterator:
-                pass
+                _()
             return
 
         # deployment and operations records are mutated by the iterator so we need to
         # commit them before iterating and at each iteration
-        session.commit()  # Update deployment status to RUNNING, operations to PENDING
-        for cluster_status_logs in deployment_iterator:
-            session.commit()  # Update operation status to RUNNING
+        for cluster_status_logs_process in deployment_iterator:
+            session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
+            cluster_status_logs = cluster_status_logs_process()
+            cluster_status_logs
             if cluster_status_logs and any(cluster_status_logs):
                 session.add_all(cluster_status_logs)
             session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -90,16 +90,16 @@ def deploy(
 
         if dry:
             for _ in deployment_iterator:
-                _()
+                if _:
+                    _()
             return
 
         # deployment and operations records are mutated by the iterator so we need to
         # commit them before iterating and at each iteration
-        session.commit()  # Update deployment status to RUNNING, operations to PENDING
+        session.commit()  # Update operation status to RUNNING
         for process_operation_fn in deployment_iterator:
             session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
-            cluster_status_logs = process_operation_fn()
-            if cluster_status_logs and any(cluster_status_logs):
+            if process_operation_fn and (cluster_status_logs := process_operation_fn()):
                 session.add_all(cluster_status_logs)
             session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
 
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -95,10 +95,10 @@ def deploy(
 
         # deployment and operations records are mutated by the iterator so we need to
         # commit them before iterating and at each iteration
-        for cluster_status_logs_process in deployment_iterator:
+        session.commit()  # Update deployment status to RUNNING, operations to PENDING
+        for process_operation_fn in deployment_iterator:
             session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
-            cluster_status_logs = cluster_status_logs_process()
-            cluster_status_logs
+            cluster_status_logs = process_operation_fn()
             if cluster_status_logs and any(cluster_status_logs):
                 session.add_all(cluster_status_logs)
             session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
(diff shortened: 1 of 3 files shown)


tdp/cli/commands/deploy.py
```diff
@@ -90,14 +90,15 @@ def deploy(
...
 
         # deployment and operations records are mutated by the iterator so we need to
         # commit them before iterating and at each iteration
-        session.commit()  # Update deployment status to RUNNING, operations to PENDING
-        for cluster_status_logs in deployment_iterator:
-            session.commit()  # Update operation status to RUNNING
+        for cluster_status_logs_process in deployment_iterator:
+            session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
+            cluster_status_logs = cluster_status_logs_process()
+            cluster_status_logs
```
PaulFarault, line 101: What is the purpose of this line?

tdp/cli/commands/deploy.py
```diff
@@ -90,14 +90,15 @@ def deploy(
...
-                pass
+                _()
             return
 
         # deployment and operations records are mutated by the iterator so we need to
         # commit them before iterating and at each iteration
-        session.commit()  # Update deployment status to RUNNING, operations to PENDING
-        for cluster_status_logs in deployment_iterator:
-            session.commit()  # Update operation status to RUNNING
+        for cluster_status_logs_process in deployment_iterator:
```
PaulFarault, line 98: You should find a better naming, sth like `process_operation_fn`.

tdp/cli/commands/deploy.py
```diff
@@ -90,14 +90,15 @@ def deploy(
 
         if dry:
             for _ in deployment_iterator:
-                pass
+                _()
             return
 
         # deployment and operations records are mutated by the iterator so we need to
         # commit them before iterating and at each iteration
-        session.commit()  # Update deployment status to RUNNING, operations to PENDING
```
PaulFarault, line 98: Why did you removed this commit? We should still save initial state to the database.

tdp/core/deployment/deployment_iterator.py
```diff
@@ -106,115 +106,122 @@ def __init__(
         except NothingToReconfigureError:
             self._reconfigure_operations = None
 
-    def __next__(self) -> Optional[list[SCHStatusLogModel]]:
+    def __next__(self) -> Callable[[], Optional[list[SCHStatusLogModel]]]:
         try:
             while True:
                 operation_rec: OperationModel = next(self._iter)
+                operation_rec.state = OperationStateEnum.RUNNING
```
PaulFarault, line 113: This is incorrect: in case of failure, operation state will be `RUNNING` and then `HELD`. IMO it should be `HELD` directly.

PaulFarault (changes requested):
There are things to improve but you're on the right track. Additional comments:
- try to set the `start_date` along with `RUNNING`
- fix the type issue

tdp/cli/commands/deploy.py
SteBaum: Since the __next__ function of the iterator class returns a callable. We have to instatiate each callable in the iterator.

tdp/cli/commands/deploy.py
SteBaum: This is done with line beneath the for loop since it commits before each iteration

tdp/core/deployment/deployment_iterator.py
SteBaum: Before the execution of the operation the state is `PENDING`. Then when its starts it is `RUNNING`. The executor changes the state from `RUNNING` to `SUCCESS` or `FAILURE` and all operations which have not been executed pass from `PENDING` to `HELD` if the deployment fails as it is stated in line 117-119.

tdp/core/deployment/deployment_iterator.py
```diff
@@ -106,115 +108,31 @@ def __init__(
...
-                if operation.name == OPERATION_SLEEP_NAME:
-                    return
-
-                sch_status_logs: list[SCHStatusLogModel] = []
-                sc_name = ServiceComponentName(
-                    service_name=operation.service_name,
-                    component_name=operation.component_name,
+                operation_rec.state = OperationStateEnum.RUNNING
+                if not operation.noop:
+                    operation_rec.start_time = datetime.utcnow()
```
PaulFarault, line 123: This is not ok, if the operation is `HELD` or `FAILURE`, it should not have a `start_date`.

In that case, you shouldn't generate the `-process_operation_fn` either.

tdp/cli/commands/deploy.py
```diff
@@ -90,14 +90,15 @@ def deploy(
...
+                _()
             return
 
         # deployment and operations records are mutated by the iterator so we need to
         # commit them before iterating and at each iteration
         session.commit()  # Update deployment status to RUNNING, operations to PENDING
-        for cluster_status_logs in deployment_iterator:
-            session.commit()  # Update operation status to RUNNING
+        for process_operation_fn in deployment_iterator:
+            session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
```
PaulFarault, line 100: No need to change this comment, the new one is false.

PaulFarault (changes requested):

tdp/core/deployment/deployment_iterator.py
PaulFarault: No, we shouldn't even start the execution if the deployment is in `FAILURE`.

PaulFarault (approved):

PaulFarault:
Should we replace:

```py
if _:
    _()
```

By

```py
_() if _ else None
```

PR #548, opened 2024-02-09
Refactor/collections init operations

PaulFarault:

Fixes None

#### Additional comments

This PR addresses the issue with assigning `collection_name` values. Previously, the assigned `collection_name` was that of the first collection in which the operation was specified in the DAG. Now, the `collection_name` for a playbook is the one in which the playbook is defined.

Another key benefit of this PR is the decoupling of the `Operation` creation logic. Playbook operations are now exclusively forged within `Collection`, and `Collections` is limited to forging Noop operations only.

I've intentionally broken down the PR into numerous commits to ease the review process. The most significant logic change is found in the "refactor: create playbook operations in Collection" commit.

#### Agreements

### DIFF
--- a/tdp/core/collections.py
+++ b/tdp/core/collections.py
@@ -172,7 +172,9 @@ class Collections(Mapping[str, Collection]):
                         )
                         # 'restart' and 'stop' operations are not defined in the DAG for
                         # noop, they need to be generated from the start operations.
-                        if read_operation.noop and "_start" in read_operation.name:
+                        if read_operation.noop and read_operation.name.endswith(
+                            "_start"
+                        ):
                             logger.debug(
                                 f"DAG Operation '{read_operation.name}' is noop, "
                                 f"creating the associated restart and stop operations."
--- a/tdp/core/collections.py
+++ b/tdp/core/collections.py
@@ -143,76 +143,106 @@ class Collections(Mapping[str, Collection]):
         dag_operations: dict[str, Operation] = {}
         other_operations: dict[str, Operation] = {}
 
-        # Init DAG Operations
         for collection in collections.values():
+            # Load DAG operations from the dag files
             for dag_file in collection.dag_yamls:
                 for read_operation in read_tdp_lib_dag_file(dag_file):
-                    if read_operation.name in dag_operations:
-                        # Merge the operation with the existing one
-                        logger.debug(
-                            f"DAG Operation '{read_operation.name}' defined in collection "
-                            f"'{dag_operations[read_operation.name].collection_name}' "
-                            f"is merged with collection '{collection.name}'"
+                    existing_operation = dag_operations.get(read_operation.name)
+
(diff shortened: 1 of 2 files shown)


SteBaum (approved):

tdp/core/collections.py
```diff
@@ -137,87 +137,115 @@ def operations(self) -> dict[str, Operation]:
...
+                    if playbook_operation := collection.playbooks.get(
+                        read_operation.name
+                    ):
+                        # Create an Operation instance based on the playbook
+                        # TODO: would be nice to dissociate the Operation class from the playbook and store the playbook in the Operation
+                        dag_operation_to_register = Operation(
+                            name=read_operation.name,
+                            collection_name=playbook_operation.collection_name,
+                            host_names=playbook_operation.host_names,
+                            depends_on=read_operation.depends_on,
```
rpignolet, line 163: `depends_on` is an array so the ref is given to the new Operation which is wrong because when using `.extend` it will update the `depends_on` for the `read_operation` too. You must use `deepcopy` for example.

Maybe it will be better to have a `clone` method for `Operation` which deep clone an operation.

tdp/core/deployment/deployment_runner.py
```diff
@@ -66,9 +66,11 @@ def _run_operation(self, operation_rec: OperationModel) -> None:
             return
 
         # Execute the operation
-        playbook_file = self._collections[operation.collection_name].playbooks[
-            operation.name
-        ]
+        playbook_file = self._collections[
+            operation.collection_name
+        ].playbooks_directory / (
+            operation.name + ".yml"
```
rpignolet, line 72: Why this modification ? This is bad, we had the path to the playbook file without this PR.

PaulFarault:
@rpignolet I rolled back the `Collection` edition to focus on `Collections` initialization. I'll improve the `Collection` class in a following PR.

rpignolet (approved):

tdp/core/collections.py
```diff
@@ -137,88 +137,120 @@ def operations(self) -> dict[str, Operation]:
...
+                        dag_operations[read_operation.name] = dag_operation_to_register
+                        continue
+
+                    # The read_operation is already registered
+                    if existing_operation:
                         logger.debug(
-                            f"DAG Operation '{operation.name}' defined in collection "
-                            f"'{self._dag_operations[operation.name].collection_name}' "
-                            f"is merged with collection '{collection_name}'"
+                            f"'{read_operation.name}' defined in "
```
SteBaum, line 183: shouldn't you write the playbook here where the operation is defined ?

PaulFarault: What do you mean?

SteBaum: sorry yes it it good

SteBaum (approved):
tested it again and it works

PaulFarault:
Force pushed after sqashing and rebasing.

PR #549, opened 2024-02-13
Refactor log messages of the logger in the TDP Core

### LINKED ISSUE #543: Refactor log messages of the logger in the TDP Core
The  messages of criticality level `INFO` should be set in the CLI and not in the Core except if the TDP Server takes exactly the same logs.

SteBaum:

Fixes #543 

#### Additional comments

- The ansible playbook command is now only displayed in mode `DEBUG` and not `INFO`.
- Iterator now also returns the operation record, which is printed with click when deployment is launched.

#### Agreements

### DIFF
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -87,10 +87,12 @@ def deploy(
         ).run(planned_deployment, force_stale_update=force_stale_update)
 
         if dry:
-            for op, fn in deployment_iterator:
-                if fn:
-                    fn()
-                click.echo(op)
+            for operation_rec, process_operation_fn in deployment_iterator:
+                if process_operation_fn:
+                    process_operation_fn()
+                click.echo(
+                    f"[DRY MODE] :Operation {operation_rec.operation} is in state {operation_rec.state}"
+                )
             return
 
         # deployment and operations records are mutated by the iterator so we need to
@@ -100,7 +102,7 @@ def deploy(
             dao.session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
             if process_operation_fn and (cluster_status_logs := process_operation_fn()):
                 click.echo(
-                    f"Operation {operation_rec.operation} is in state {operation_rec.state}"
+                    f"Operation {operation_rec.operation} is in state {operation_rec.state} {'in hosts :' + operation_rec.host if operation_rec.host is not None else ''}"
                 )
                 dao.session.add_all(cluster_status_logs)
             dao.session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -91,7 +91,7 @@ def deploy(
                 if process_operation_fn:
                     process_operation_fn()
                 click.echo(
-                    f"[DRY MODE] :Operation {operation_rec.operation} is {operation_rec.state}"
+                    f"[DRY MODE]: Operation {operation_rec.operation} is {operation_rec.state}"
                 )
             return
 
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -89,7 +89,7 @@ def deploy(
         if dry:
             for op, fn in deployment_iterator:
                 if fn:
-                    fn(op)
+                    fn()
                 click.echo(op)
             return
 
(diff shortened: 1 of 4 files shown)


PaulFarault (approved):

tdp/cli/commands/deploy.py
```diff
@@ -89,17 +89,21 @@ def deploy(
...
         # commit them before iterating and at each iteration
         session.commit()  # Update operation status to RUNNING
-        for process_operation_fn in deployment_iterator:
+        for operation_rec, process_operation_fn in deployment_iterator:
             session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
-            if process_operation_fn and (cluster_status_logs := process_operation_fn()):
+            if process_operation_fn and (
+                cluster_status_logs := process_operation_fn(operation_rec)
+            ):
+                click.echo(operation_rec)
```
PaulFarault, line 106: We may want to display a meaningful message here.

tdp/core/deployment/deployment_iterator.py
```diff
@@ -137,7 +135,9 @@ def __next__(
             self.deployment.state = DeploymentStateEnum.FAILURE
             raise e
 
-    def _process_operation_fn(self, operation_rec) -> Optional[list[SCHStatusLogModel]]:
+    def _process_operation_fn(
+        self, operation_rec: OperationModel
```
rpignolet, line 139: Why there is a `operation_rec` parameter ?

rpignolet: The parameter here is ok, the problem is the partial above.

tdp/core/deployment/deployment_iterator.py
```diff
@@ -111,20 +110,19 @@ def __init__(
...
                 if self.deployment.state == DeploymentStateEnum.FAILURE:
                     operation_rec.state = OperationStateEnum.HELD
-                    return
+                    return operation_rec, None
 
                 operation_rec.state = OperationStateEnum.RUNNING
 
-                return partial(self._process_operation_fn, operation_rec)
-
+                return operation_rec, self._process_operation_fn
```
rpignolet, line 125: Why removing the `partial` ?

PaulFarault: It seems better if the caller can pass the operation as an argument directly instead of pre-passing it via partial.

rpignolet: No the caller can give his own operation which he must not be able to. Only the operation provided by the iterator must be ran.

tdp/cli/commands/deploy.py
SteBaum: operation_rec already prints so much parameters which the user can check afterwards in the DB and adding extra texts will print to much output in terminal.

I propose only displaying the name of the operation and its state.
```py
f"Operation {operation_rec.operation} is in state {operation_rec.state}"
```

tdp/core/deployment/deployment_iterator.py
SteBaum: I reverted it back to the previous version so that the operation_rec of the iterator is directly integrated.

tdp/core/deployment/test_deployment_runner.py
```diff
@@ -111,8 +111,10 @@ def test_deployment_plan_with_filter_is_success(
     deployment_iterator = mock_deployment_runner.run(deployment)
 
     for i, _ in enumerate(deployment_iterator):
```
rpignolet, line 113: Syntaxe like `for i, (op, fn) in enumerate(deployment_iterator):` does not work ?

tdp/core/deployment/test_deployment_runner.py
```diff
@@ -204,9 +207,12 @@ def test_service_log_not_emitted_when_config_start_wrong_order(
     )
     deployment_iterator = mock_deployment_runner.run(deployment)
 
-    for _ in deployment_iterator:
-        if _:
-            _()
+    for (
+        op,
+        fn,
+    ) in deployment_iterator:
```
rpignolet, line 213: You can not one line this like other ? This is enforce by the linter ?

SteBaum: No it was not enforeced, it has been written on one line

tdp/core/deployment/test_deployment_runner.py
SteBaum: Sorry didn't see this comment before. yes it works, it has been rewritten

PaulFarault:
Some conflicts must be resolved before merging.

tdp/cli/commands/deploy.py
```diff
@@ -87,17 +87,21 @@ def deploy(
         ).run(planned_deployment, force_stale_update=force_stale_update)
 
         if dry:
-            for _ in deployment_iterator:
-                if _:
-                    _()
+            for op, fn in deployment_iterator:
```
PaulFarault, line 90: Let's give `fn` a meaningful name and use the same everywhere.

PaulFarault, line 90: Let's give `fn` a meaningful name and use the same everywhere.

PaulFarault (changes requested):

tdp/cli/commands/deploy.py
```diff
@@ -87,17 +87,21 @@ def deploy(
...
+                if fn:
+                    fn()
+                click.echo(op)
```
PaulFarault, line 93: Maybe add a "[DRY MODE]" prefix.

PaulFarault (approved):

rpignolet (approved):

PR #550, opened 2024-02-14
513 prevent overriding existing plan

### LINKED ISSUE #513: Prevent overriding an existing plan
### Issue

Current behavior when creating a new deployment plan - using the `tpd plan [dag|ops|resume|reconfigure]` commands - is to replace any existing deployment plan by the new one. This could lead to several issues:

- User incomprehension, as they are not informed that any existing plan will be lost.
- Unexpected database behavior, as saving the new operations is for now considered as an update. Any field that is not overwritten by the new operation will keep the previous one value. This lead us to explicitly set `None` values for any unused column (for [ex](https://github.com/TOSIT-IO/tdp-lib/pull/505#discussion_r1397087362)).

### Suggestion

I suggest:

- Add a `[Y/n]` confirmation prompt along with a `--force` option to the generation commands.
- Erasing the existing deployment plan before creating a new one.

### Discussion

Mutations for planned deployments have been chosen in order to keep a consistent id increment between deployments: We ensure that if `n` is the last deployment, the next one will be `n + 1` no matter how many plan have been made.

Should we keep this incrementation?

SteBaum:

Fixes #513 

#### Additional comments

- Added `y/N` confirmation to override existing deployment plan in plan dag, reconfigure, import and ops.
- Added `--force` option to automatically override existing deployment plan without asking the question.

#### Agreements

### DIFF
--- a/tdp/cli/commands/plan/dag.py
+++ b/tdp/cli/commands/plan/dag.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Optional
 import click
 
 from tdp.cli.params import collections_option, database_dsn_option
-from tdp.cli.params.plan import preview_option, rolling_interval_option
+from tdp.cli.params.plan import force_option, preview_option, rolling_interval_option
 from tdp.cli.queries import get_planned_deployment
 from tdp.cli.utils import print_deployment
 from tdp.core.dag import Dag
@@ -69,6 +69,7 @@ def _validate_filtertype(
 )
 @rolling_interval_option
 @preview_option
+@force_option
 @collections_option
 @database_dsn_option
 def dag(
@@ -76,6 +77,7 @@ def dag(
     targets: tuple[str],
     restart: bool,
     preview: bool,
+    force: bool,
     collections: Collections,
     database_dsn: str,
     reverse: bool,
@@ -120,7 +122,7 @@ def dag(
     with Dao(database_dsn, commit_on_exit=True) as dao:
         planned_deployment = get_planned_deployment(dao.session)
         if planned_deployment:
-            if click.confirm(
+            if force or click.confirm(
                 "A deployment plan already exists, do you want to override it?"
             ):
                 deployment.id = planned_deployment.id
--- a/tdp/cli/commands/plan/dag.py
+++ b/tdp/cli/commands/plan/dag.py
@@ -120,6 +120,12 @@ def dag(
     with Dao(database_dsn, commit_on_exit=True) as dao:
         planned_deployment = get_planned_deployment(dao.session)
         if planned_deployment:
-            deployment.id = planned_deployment.id
+            if click.confirm(
+                "A deployment plan already exists, do you want to override it?"
+            ):
+                deployment.id = planned_deployment.id
+            else:
+                click.echo("No new deployment plan has been created.")
+                return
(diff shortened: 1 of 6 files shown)


tdp/cli/commands/plan/dag.py
```diff
@@ -70,6 +70,11 @@ def _validate_filtertype(
     is_flag=True,
     help="Replace 'start' operations by 'stop' operations. This option should be used with `--reverse`.",
 )
+@click.option(
+    "--force",
+    is_flag=True,
+    help="Automatically overrides existing deployment plan.",
```
PaulFarault, line 76:
```suggestion
    help="Force overriding an existing deployment plan.",
```

tdp/cli/commands/plan/import_file.py
```diff
@@ -42,7 +48,19 @@ def import_file(
             )
             # if a planned deployment is present, update it instead of creating it
```
PaulFarault, line 49: Update comments

PaulFarault (changes requested):
Nice PR! Once finished, try to think if we should, and if yes how, refactor the checking logic to keep it in a single place.

PaulFarault (approved):

tdp/cli/commands/plan/dag.py
```diff
@@ -124,6 +130,14 @@ def dag(
...
+                "A deployment plan already exists, do you want to override it?"
+            )
+            if confirmed:
+                dao.session.delete(planned_deployment)
+                click.echo("Previous planned deployment has been overritten")
+            else:
+                click.echo("No new deployment plan has been created")
+                return
         dao.session.merge(deployment)
     click.echo("Deployment plan successfully created.")
```
PaulFarault, line 143: I suggest this instead so that it better stick to the current implementation:

```suggestion
            if confirmed := click.confirm("A deployment plan already exists, do you want to override it?"):
                deployment.id = planned_deployment.id
            else:
                click.echo("No new deployment plan has been created.")
                return
    dao.session.merge(deployment)
    click.echo("Deployment plan successfully created.")
```

If that's good for you, please propagate it to the other command.

PaulFarault (changes requested):

tdp/cli/commands/plan/dag.py
```diff
@@ -71,6 +71,11 @@ def _validate_filtertype(
     is_flag=True,
     help="Replace 'start' operations by 'stop' operations. This option should be used with `--reverse`.",
 )
+@click.option(
+    "--force",
+    is_flag=True,
+    help="Force overriding an existing deployment plan.",
+)
```
PaulFarault, line 78: Refactor the option so that it can be reused.

PaulFarault (changes requested):

tdp/cli/commands/plan/dag.py
SteBaum: implemented almost the same, just removed `confirmed := ` on line 133 and equivalent for other files.

PaulFarault (approved):

rpignolet (commented):
We should introduce a global option `--batch-mode` like Maven in order to prevent any prompt, can you make a PR for it ?

rpignolet (approved):

PR #554, opened 2024-02-20
feat: query get_operation_records prints correct errors

### LINKED ISSUE #483: `tdp browse` prints wrong info with wrong parameters
Multiple issues:

1. When displaying an operation of unexisting deployment:

```
➜ tdp browse 1000 not_found
Operation(s) details
```

It must error with:

```
Deployment id 14346 does not exist.
```

2. When displaying an unexisting operation of existing deployment:

```
➜ tdp browse 1 not_found                 
Operation(s) details
```

It must error with:

```
Operation "not_found" does not exist in deployment 1.
```

SteBaum:

Fixes #483 

#### Additional comments

Modified the error handling to print the right messages

#### Agreements

### DIFF
--- a/tdp/cli/queries.py
+++ b/tdp/cli/queries.py
@@ -228,6 +228,6 @@ def get_operation_records(
         return query_result
 
     else:
-        raise Exception(
+        raise NoResultFound(
             f"Operation {operation_name} does not exist in deployment {deployment_id} or deployment {deployment_id} does not exist."
         )
--- a/tdp/cli/queries.py
+++ b/tdp/cli/queries.py
@@ -217,15 +217,17 @@ def get_operation_records(
         List of matching operation records.
 
     Raises:
-        NoResultFound: If the operation does not exist.
+        NoResultFound: If the operation or deployment does not exist.
     """
-    try:
-        return (
-            session.query(OperationModel)
-            .filter_by(deployment_id=deployment_id, operation=operation_name)
-            .all()
-        )
-    except NoResultFound as e:
+    query_result = (
+        session.query(OperationModel)
+        .filter_by(deployment_id=deployment_id, operation=operation_name)
+        .all()
+    )
+    if query_result:
+        return query_result
+
+    else:
         raise Exception(
-            f"Operation {operation_name} does not exist in deployment {deployment_id}."
-        ) from e
+            f"Operation {operation_name} does not exist in deployment {deployment_id} or deployment {deployment_id} does not exist."
+        )


PaulFarault:
This PR defines 2 queries in one. It makes code way more complicated and does not provide much value.

Checking if the deployment exist may be a good idea but it should be done in it's own query.

tdp/cli/queries.py
```diff
@@ -449,7 +449,7 @@ def get_planned_deployment(session: Session) -> Optional[DeploymentModel]:
 
 def get_operation_records(
     session: Session, deployment_id: int, operation_name: str
-) -> list[OperationModel]:
+) -> list[OperationModel] | None:
```
PaulFarault, line 452: `|` is not available with Python 3.9

PaulFarault (changes requested):
Focus on resolving the issue mentioned by @sergkudinov on this PR:

> When you changed to `.all()` [here](https://github.com/TOSIT-IO/tdp-lib/commit/a3060cb5df9e098d6c0e0ec24961b829c3ccfeeb#diff-6eb19396e554b1abab6190572020f2cb35d5b3c8f7bd0fa91b697a99fc26e1cbR438-R440), the `NoResultFound` exception is never raised.

Then, we could study how to raise a more specific Error. 

As now we're doing 3 queries which seems a lot.

SteBaum:
Opted for a very simple change, to through an error if no results are found and modified the message to incorporate the two possibilities.

tdp/cli/queries.py
```diff
@@ -217,15 +217,17 @@ def get_operation_records(
...
+    )
+    if query_result:
+        return query_result
+
+    else:
         raise Exception(
-            f"Operation {operation_name} does not exist in deployment {deployment_id}."
-        ) from e
+            f"Operation {operation_name} does not exist in deployment {deployment_id} or deployment {deployment_id} does not exist."
+        )
```
PaulFarault, line 233: You can even be more generic:

```suggestion
            f"Can't find operation {operation_name} for deployment {deployment_id}."
        ) from e
```

PaulFarault (changes requested):

tdp/cli/queries.py
SteBaum: The whole point of the prolonged message was to state the two possible issues with the query, either the operation does not exist in the chosen deployment or the deployment itself does not exist. this message is more specific.
However I can change above the `raise Exception` to  `raise NoResultsFound` which is also more specific.

PR #555, opened 2024-02-20
Initialize `Playbook` entities inside of `Collection`

PaulFarault:

Fixes None

#### Additional comments

Change the `Collection.playbooks` properties so that it maps operation names to `Playbook` object instead of playbook Paths.

In another PR, we should remove hosts from `Operation` definition so that host are only stored in `Playbook`.

#### Agreements


SteBaum (approved):
Reviewed and tested

rpignolet (approved):

PR #579, opened 2024-04-02
`--database-dsn` option returns directly an engine

PaulFarault:

Fixes None

#### Additional comments

- Add a callback to the `--database-dsn` option to return an engine directly
- Make the `Dao` class take an engine as an input instead of a database DSN

#### Agreements


tdp/cli/commands/test_init.py
```diff
@@ -5,16 +5,17 @@
...
 
 
-def test_tdp_init(collection_path: Path, database_dsn: str, vars: Path):
+def test_tdp_init(collection_path: Path, database_engine: Engine, vars: Path):
     args = [
         "--collection-path",
         collection_path,
         "--database-dsn",
-        database_dsn,
+        database_engine,
```
rpignolet, line 18: I think, the arg should be a `str` no ?

PaulFarault: You're right, I mixed it with the `test_dag_order` tests. I fixed it, thanks!

rpignolet (approved):

SteBaum:
The logs have become unreadable, the queries are returned in the logs and `tdp deploy --mock-deploy`does not return the message that the deployment was a success at the end (or it is hidden in the unreadable logs). I needed to do a `tdp browse` to check.

tdp/cli/params/database_dsn_option.py
```diff
@@ -3,18 +3,35 @@
...
+
+    Args:
+        ctx: Click context.
+        param: Click parameter.
+        value: Database DSN.
+
+    Returns:
+        Engine from the DSN.
+    """
+    return create_engine(value, echo=True, future=True)
```
SteBaum, line 20: this `echo=True` pollutes the logs printed in the terminal.

SteBaum (approved):

PR #563, opened 2024-02-28
Dev/alembic migration

SteBaum:
#### Which issue(s) this PR fixes

None

#### Additional comments

- Added Alembic in `pyproject.toml` and did the setup to use the metadata of the models for MySQL, Postgres and SQLite
- Added the use Alembic in the `alembic_migration/README.md`.

#### Agreements

- [x] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0).
- [x] I hereby agree to grant [TOSIT](https://www.tosit.io/) a copyright license to use my contributions.

PR #584, opened 2024-05-02
refactor: engine in test procedure

SteBaum:

Fixes None

#### Additional comments

This PR is necessary for the PR Dev/alembic migration #563. When we integrate Alembic in the `init` command, it creates a new engine and synchronization problems occur between the engines when launching the unit tests although it works manually when we perform one command after another. Therefore, the `init` command has been left out of all unit tests and replaced with hard coded initialization except the test which specifically tests the `init` command of course.

The engine is created and closed also in the tests since we import the functions with the `click` decorators which take as input a `database-dsn` and not an engine.

#### Agreements

### DIFF
--- a/tdp/cli/commands/conftest.py
+++ b/tdp/cli/commands/conftest.py
@@ -1,11 +1,37 @@
 # Copyright 2022 TOSIT.IO
 # SPDX-License-Identifier: Apache-2.0
 
+from collections.abc import Generator
 from pathlib import Path
+from typing import Any
 
 import pytest
+from click.testing import CliRunner
+from sqlalchemy import create_engine
 
+from tdp.cli.commands.init import init
 from tdp.conftest import generate_collection_at_path
+from tdp.core.models import BaseModel
+
+
+@pytest.fixture
+def tdp_init(
+    collection_path: Path, db_dsn: str, vars: Path
+) -> Generator[list[Any], Any, None]:
+    base_args = [
+        "--collection-path",
+        collection_path,
+        "--database-dsn",
+        db_dsn,
+        "--vars",
+        vars,
+    ]
+    runner = CliRunner()
+    runner.invoke(init, base_args)
+    yield base_args
+    engine = create_engine(db_dsn)
+    BaseModel.metadata.drop_all(engine)
+    engine.dispose()
 
 
 @pytest.fixture
--- a/tdp/cli/commands/conftest.py
+++ b/tdp/cli/commands/conftest.py
@@ -1,6 +1,7 @@
 # Copyright 2022 TOSIT.IO
 # SPDX-License-Identifier: Apache-2.0
 
+from collections import namedtuple
 from collections.abc import Generator
 from pathlib import Path
 from typing import Any
@@ -13,12 +14,14 @@ from tdp.cli.commands.init import init
 from tdp.conftest import generate_collection_at_path
 from tdp.core.models import BaseModel
 
+tdp_init_args = namedtuple("tdp_init_args", ["collection_path", "db_dsn", "vars"])
+
 
 @pytest.fixture
 def tdp_init(
     collection_path: Path, db_dsn: str, vars: Path
-) -> Generator[tuple[Path, str, Path], Any, None]:
-    tdp_init_args = [
+) -> Generator[tdp_init_args, Any, None]:
+    base_args = [
         "--collection-path",
         collection_path,
         "--database-dsn",
@@ -27,8 +30,8 @@ def tdp_init(
         vars,
     ]
     runner = CliRunner()
-    runner.invoke(init, tdp_init_args)
-    yield (collection_path, db_dsn, vars)
+    runner.invoke(init, base_args)
+    yield tdp_init_args(collection_path, db_dsn, vars)
     engine = create_engine(db_dsn)
     BaseModel.metadata.drop_all(engine)
     engine.dispose()
--- a/tdp/cli/commands/conftest.py
+++ b/tdp/cli/commands/conftest.py
@@ -17,8 +17,8 @@ from tdp.core.models import BaseModel
 @pytest.fixture
 def tdp_init(
     collection_path: Path, db_dsn: str, vars: Path
-) -> Generator[list[Any], Any, None]:
-    base_args = [
+) -> Generator[tuple[Path, str, Path], Any, None]:
+    tdp_init_args = [
         "--collection-path",
         collection_path,
         "--database-dsn",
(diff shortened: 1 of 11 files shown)


tdp/cli/commands/plan/test_dag.py
```diff
@@ -5,20 +5,34 @@
...
+):
     base_args = [
         "--collection-path",
         collection_path,
         "--database-dsn",
         db_dsn,
     ]
+    # Create engine and initialize the database
+    engine = create_engine(db_dsn)
+    init_database(engine)
```
rpignolet, line 29: You must not init the database like this. If the `assert` below or any lines of code raise an exception, the `drop_all` code will not run.

SteBaum: It has been corrected with the second commit, However, now the engine is created twice, once through the fixture and once through the decorator `@database_dsn_option` which accompanies each cli function.

tdp/cli/commands/conftest.py
```diff
@@ -31,3 +35,15 @@ def collection_path(tmp_path_factory: pytest.TempPathFactory) -> Path:
 @pytest.fixture
 def vars(tmp_path_factory: pytest.TempPathFactory) -> Path:
     return tmp_path_factory.mktemp("collection")
+
+
+@pytest.fixture
+def create_cluster_variables(collection_path, vars) -> ClusterVariables:
```
rpignolet, line 41: Other fixtures do not have a verb, you should rename it `cluster_variables`.
```suggestion
def cluster_variables(collection_path, vars) -> ClusterVariables:
```

SteBaum: done

tdp/cli/commands/conftest.py
```diff
@@ -1,11 +1,35 @@
...
+        "--database-dsn",
+        db_dsn,
+        "--vars",
+        vars,
+    ]
+    runner = CliRunner()
+    yield runner.invoke(init, args)
+    engine = create_engine(db_dsn)
+    BaseModel.metadata.drop_all(engine)
+    engine.dispose()
```
PaulFarault, line 32: This should run `tdp init` and then yield `collection-path`, `database-dsn` and `vars` (as a named tuple or directly as an usable table like `args` in the current fixture).

PaulFarault (changes requested):

tdp/cli/commands/conftest.py
SteBaum: Has been changed, a bit complicated with the typing of the fixture.

tdp/cli/commands/conftest.py
```diff
@@ -1,11 +1,37 @@
...
+def tdp_init(
+    collection_path: Path, db_dsn: str, vars: Path
+) -> Generator[list[Any], Any, None]:
+    base_args = [
+        "--collection-path",
+        collection_path,
+        "--database-dsn",
+        db_dsn,
+        "--vars",
+        str(vars),
```
PaulFarault, line 27: Why casting only the `vars` while `collection_path` is also a Path?

Either don't cast any or cast both to return a `list[str]`

SteBaum: Because we mostly casted the `vars` before while not casting the `collection_path`. 
However, it works if we do not cast any. I do not know the reason why the `vars` were casted before in some tests, I though there was a specific reason. I will just remove it.

tdp/cli/commands/plan/test_ops.py
```diff
@@ -1,23 +1,15 @@
...
-        db_dsn,
-    ]
+def test_tdp_plan_run(
+    tdp_init: list,
+):
     runner = CliRunner()
-    result = runner.invoke(init, [*base_args, "--vars", str(vars)])
-    assert result.exit_code == 0, result.output
-    result = runner.invoke(ops, [*base_args, "service_install"])
+    result = runner.invoke(ops, [*tdp_init[:-2], "service_install"])
```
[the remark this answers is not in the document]
SteBaum: If we pass it it, we have to exclude it again for every other tests which uses `tdp_init`

tdp/cli/commands/plan/test_dag.py
```diff
@@ -2,23 +2,20 @@
...
 
-def test_tdp_plan_dag(collection_path: Path, db_dsn: str, vars: Path):
-    base_args = [
+def test_tdp_plan_dag(
+    tdp_init: tuple,
+):
+    tdp_init_args = [
         "--collection-path",
-        collection_path,
+        tdp_init[0],
```
PaulFarault, line 15: Would be better with a named tuple.

PaulFarault (changes requested):

tdp/cli/commands/plan/test_dag.py
SteBaum: done

tdp/cli/commands/plan/test_dag.py
```diff
@@ -2,23 +2,21 @@
...
+    tdp_init: tdp_init_args,
+):
     base_args = [
         "--collection-path",
-        collection_path,
+        tdp_init.collection_path,
         "--database-dsn",
-        db_dsn,
+        tdp_init.db_dsn,
     ]
```
PaulFarault, line 19: No need for those intermediary arrays as only one command is executed from now.

PaulFarault (changes requested):

tdp/cli/commands/plan/test_dag.py
SteBaum: done

tdp/cli/commands/plan/test_ops.py
```diff
@@ -1,23 +1,27 @@
...
-    result = runner.invoke(ops, [*base_args, "service_install"])
+    result = runner.invoke(
+        ops,
+        [
+            *[
+                "--collection-path",
+                tdp_init.collection_path,
+                "--database-dsn",
+                tdp_init.db_dsn,
+            ],
```
PaulFarault, line 23: No need to create a list in this situation.

PaulFarault (approved):

tdp/cli/commands/conftest.py
```diff
@@ -1,11 +1,43 @@
...
+    vars: Path
+
+
+@pytest.fixture
+def tdp_init(
+    collection_path: Path, db_dsn: str, vars: Path
+) -> Generator[TDPInitArgs, None, None]:
+    base_args = [
+        "--collection-path",
+        collection_path,
```
rpignolet, line 29: `collection_path` is a `Path`, we should not have a list with multiple type for element, use `str(collection_path)` to convert `Path` to `str`. Same for all occurrences.

SteBaum: done

tdp/conftest.py
```diff
@@ -72,6 +68,7 @@ def db_engine(
     if request.param:
         init_database(engine)
     yield engine
+    BaseModel.metadata.drop_all(engine)
```
rpignolet, line 71: We init the database only if `request.param` is `True`, so we must `drop_all` only if `request.param` is `True`.
```suggestion
    if request.param:
        BaseModel.metadata.drop_all(engine)
```

SteBaum: done

rpignolet (approved):

PR #589, opened 2024-05-16
fix: set byte length for operation.logs column

SteBaum:
#### Which issue(s) this PR fixes

Fixes None

#### Additional comments

Deployment on MariaDB fails if no byte length is set for the column `logs` in table `operation`. The length has been set at `10 000 000`. It has been tested with `1 000 000` and it worked. The maximum length in a deployment is `grafana_server_config   | 611501`.

In PR #590 a test is included which adds text in the logs column to see if passes on every database.

#### Agreements

- [x] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0).
- [x] I hereby agree to grant [TOSIT](https://www.tosit.io/) a copyright license to use my contributions.

PR #590, opened 2024-05-16
feat: added test for length of logs in operation table

SteBaum:

Fixes None

#### Additional comments

Goes with  `fix: set byte length for operation.logs column` #589 as a test which checks if the `logs` column in the table `operation` can handle a large amount of bytes on each databases. The amount of characters has been set to `10 000 000`. Before the PR #589 the test failed on `MySQL` and `MariaDB` and with the PR it succeeds on these databases.

#### Agreements


SteBaum:
Just added the command for the extra test in the documentation with the last force push

poetry.lock
rpignolet, line 1: Did you use `poetry lock --no-update` to update this file ? You add one dependency and there are 14 dependencies updated.

SteBaum: No I will correct this

SteBaum: corrected with force push

pyproject.toml
```diff
@@ -29,6 +29,7 @@ pydantic = "^2.6.0"
 psycopg2 = { version = "^2.9.9", optional = true }
 psycopg2-binary = { version = "^2.9.9", optional = true }
 pymysql = { version = "^1.1.0", optional = true }
+python-lorem = "^1.3.0.post1"
```
rpignolet, line 32: This is a dependency for running test so it should be in `tool.poetry.group.dev.dependencies` section.

rpignolet (changes requested):

pyproject.toml
SteBaum: done

rpignolet (approved):

PaulFarault (approved):

PR #591, opened 2024-05-24
feat: added the option to read the --force option from .env file

SteBaum:

Fixes None

#### Additional comments

This PR is related to  PR 513 prevent overriding existing plan #550 which is already merged. It enables the user to have an environment variable `FORCE_PLAN` which if set to `True` which will automatically force the overriding of the previous plan without the prompt asking for confirmation so it disables the interactive mode and no need to put the option `--force` each time after the `tdp plan` command and acts as a batch-mode.

#### Agreements


tdp/cli/params/plan/force_option.py
```diff
@@ -12,6 +12,8 @@ def force_option(func: FC) -> FC:
     """
     return click.option(
         "--force",
+        envvar="FORCE_PLAN",
         is_flag=True,
+        default=False,
```
PaulFarault, line 17: `False` is already default.

tdp/cli/params/plan/force_option.py
```diff
@@ -12,6 +12,8 @@ def force_option(func: FC) -> FC:
     """
     return click.option(
         "--force",
+        envvar="FORCE_PLAN",
```
PaulFarault, line 15: All other tdp env vars are prefixed by `TDP_`. The naming is also a bit confusing.

```suggestion
        envvar="TDP_FORCE_PLAN_OVERRIDE",
```

PaulFarault (changes requested):

tdp/cli/params/plan/force_option.py
SteBaum: yes but there it is explicit

PaulFarault (approved):

tdp/cli/params/plan/force_option.py
SteBaum: anyways changed it

rpignolet (approved):

PR #592, opened 2024-05-29
refactor: integrated queries in dao and added tests

SteBaum:

Fixes None

#### Additional comments

This PR has integrated all queries in the DAO. The minimum transformation has been done for it to work, however, it is not finished as both ORM models and the DAO object classes are transformed several times from one to the other which does not make it efficient. For instance it is first an ORM to make the dag or for the DeploymentIterator later transformed to a DAO and then again to an ORM for the display so it cannot stay like that. The work must continue and transform all ORMs to DAO objects except when commited to the database.

Additionally tests of the DAO and DAO objects have been integrated.

#### Agreements

### DIFF
--- a/tdp/cli/commands/browse.py
+++ b/tdp/cli/commands/browse.py
@@ -8,18 +8,13 @@ import click
 from sqlalchemy import Engine
 
 from tdp.cli.params import database_dsn_option
-from tdp.cli.queries import (
-    get_deployment,
-    get_deployments,
-    get_operation_records,
-    get_planned_deployment,
-)
 from tdp.cli.utils import (
     print_deployment,
     print_object,
     print_table,
 )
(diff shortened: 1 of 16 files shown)


PaulFarault:
I agree. The result is too complicated to be propagated for all current and future queries.

However, a lot of things happen in this PR. I can identify 2 challenges to tackle here:

1. Moving create, update and delete queries to the DAO
1. Testing the queries (for both read and altering queries)

I suggest we convert this PR as draft and split this work into smaller incremental PRs to try to find a solution.

The `get_deployments` query seems to be a good one for trying to add tests. Let's try to start there by first adding it to the DAO and then adding tests to it.

PaulFarault:
IMO the new entities are not needed and we can directly use SQLAlchemy models in most cases.

Note: creating entities for status was needed are those were not directly part of our model.

PR #602, opened 2024-06-11
Stephan/refactor/alembic init

SteBaum:

Fixes None

#### Additional comments

Integrates `Alembic` in the `tdp init` command. If the database has already got an `alembic_version` table with a revision id it upgrades the tables, otherwise it creates them and puts the stamp of the current revision.

This PR includes all initial revision scripts.

#### Agreements


tdp/cli/commands/init.py
```diff
@@ -31,19 +36,49 @@
     multiple=True,
     help="Path to TDP variables overrides. Can be used multiple times. Last one takes precedence.",
 )
+@click.argument(
+    "alembic_ini_path",
+    envvar="ALEMBIC_CONFIG",
```
PaulFarault, line 41: Why not add the `TDP_` prefix for this one?

tdp/cli/commands/init.py
```diff
@@ -31,19 +36,49 @@
...
+    db_engine: str,
     validate: bool,
     vars: Path,
 ):
     """Initialize the database and the TDP variables."""
-    init_database(db_engine)
+    engine = create_engine(db_engine)
+    with engine.connect() as connection:
+        alembic_config = Config(
+            file_=str(alembic_ini_path),
```
PaulFarault, line 61: Why setting `path_type=Path` for the option if the goal is to cast it as a string right after?

tdp/cli/commands/init.py
```diff
@@ -31,19 +36,49 @@
...
-@database_dsn_option
+@database_dsn_option(create_engine=False)
 @validate_option
 @vars_option(exists=False)
 def init(
     overrides: tuple[Path],
     collections: Collections,
-    db_engine: Engine,
+    alembic_ini_path: Path,
+    db_engine: str,
```
PaulFarault, line 53: This parameter should be `database_dsn` if the `create_engine=False` is passed to the `@database_dsn_option`.

PaulFarault (changes requested):
This means that we want to use the `tdp init` command for both initialization AND migration?

tdp/cli/commands/init.py
SteBaum: Resolved it another way. Gave the absolute path of the tdp-lib folder

tdp/cli/commands/init.py
```diff
@@ -32,18 +37,42 @@
...
-    db_engine: Engine,
+    database_dsn: str,
     validate: bool,
     vars: Path,
 ):
     """Initialize the database and the TDP variables."""
-    init_database(db_engine)
+    engine = create_engine(database_dsn)
+    with engine.connect() as connection:
+        tdp_lib_folder_path: Path = Path(__file__).parents[3]
```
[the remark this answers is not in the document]
SteBaum: Fonction has been readapted but integrated with the last force push.

tdp/cli/commands/init.py
```diff
@@ -32,18 +37,42 @@
...
+        alembic_config = Config(
+            file_=f"{tdp_lib_folder_path}/alembic.ini",
```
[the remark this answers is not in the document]
SteBaum: Done

SteBaum:
`alembic_migration` folder and `alembic.ini` have been moved into the folder `tdp` so that they are included in the packaging.

PR #603, opened 2024-06-25
fix: numpy version

SteBaum:

Fixes None

#### Additional comments

Fixes numpy version compatibility with matplotlib when installing with pip.

#### Agreements

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2082,9 +2082,9 @@ docs = ["Sphinx", "sphinx-rtd-theme"]
 mysql = ["pymysql"]
 postgresql = ["psycopg2"]
 postgresql-binary = ["psycopg2-binary"]
-visualization = ["matplotlib", "pydot"]
+visualization = ["matplotlib", "numpy", "pydot"]
 
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.9.0,<4.0"
-content-hash = "8a9fdf475d30db15a31398d8744ebe59763cb88af69c73833c6a168d6deaf265"
+content-hash = "f04fe6ff12281ce991908759ef5152842bca29e909aaca46f10cc2be2267e8d9"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -31,6 +31,7 @@ psycopg2-binary = { version = "^2.9.9", optional = true }
 pymysql = { version = "^1.1.0", optional = true }
 alembic = "^1.13.1"
 alembic-postgresql-enum = "^1.2.0"
+numpy = "<2.0.0"
```
rpignolet, line 34: This is an optional dependency and enable only with `visualization` extras. `numpy` is used by `matplotlib` but when installed, `networkx` will use it and need a version `<2.0.0`.

poetry.lock
```diff
@@ -968,47 +968,47 @@ scipy = ["scipy"]
 
 [[package]]
 name = "numpy"
-version = "1.26.3"
+version = "1.26.4"
```
rpignolet, line 971: Why `numpy` is upgraded ? Did you run `poetry lock --no-update` ?

SteBaum: done

rpignolet: I do not see any change.

SteBaum: yes I did a `poetry lock --no-update` but when I add numpy it with a poetry command it upgrades it. I added it manually and now it should be good.

rpignolet (approved):

PaulFarault:
For reference, this was not a compatibility issue between `matplotlib` and `numpy` but between `networkx` and `numpy`.

`networkx` `2.5.1` is not compatible with `numpy` `2.0`. The following is enough to make it break:

```sh
pip install 'networkx[numpy]==2.5.1'
```

`numpy` is an optional dependency for `networkx` but required by `matplotlib`. Hence, installing `matplotlib` through `pip install .[visualization]` will break `networkx`.

The issue has been solved from `networkx` 3.3: https://github.com/networkx/networkx/pull/7390

PR #605, opened 2024-06-27
fix: change value order for is_active column when using query _create_last_value_statement

SteBaum:

Fixes None

#### Additional comments

The function `_create_get_sch_latest_status_statement` in `tdp/dao.py` used by the query from the command `tdp status show` takes the `latest_is_active`, and until now `None` meant True for `is_active`. However, in the order of the query False precedes None so the query will first return a False before a None which is not the expected behavior. Therefore, the order has been changed for the is_active column.

#### Agreements


SteBaum:
is_active column has been removed

PR #606, opened 2024-06-27
fix: to_restart and to_config in get_hosted_entity_statuses

SteBaum:

Fixes None

The `to_restart` and `to_config` indicators must be true only if component is active, otherwise if it belongs to a node that has been decommissioned there is no point to restart or reconfigure it and it will fail anyways since the host is not there anymore. Therefore an additional condition has been put for these indicators to be true.

#### Agreements

### DIFF
--- a/tdp/cli/commands/plan/reconfigure.py
+++ b/tdp/cli/commands/plan/reconfigure.py
@@ -39,7 +39,7 @@ def reconfigure(
         deployment = DeploymentModel.from_stale_hosted_entities(
             collections=collections,
             stale_hosted_entity_statuses=dao.get_hosted_entity_statuses(
-                filter_stale=True
+                filter_stale=True, filter_active=True
             ),
             rolling_interval=rolling_interval,
         )


PaulFarault:
Seems ok, please add a comment and/or something in the docstring.

SteBaum:
> Seems ok, please add a comment and/or something in the docstring.

Has been done.

PaulFarault:
Well, I spoke too soon. Why make the reading of the `to_restart` and `to_config` columns depend on the `is_active` column?

Can't we just use the `filter_active` filter to only select "truly" configurable/restartable components instead of altering values on read?

SteBaum:
> Well, I spoke too soon. Why make the reading of the `to_restart` and `to_config` columns depend on the `is_active` column?
> 
> Can't we just use the `filter_active` filter to only select "truly" configurable/restartable components instead of altering values on read?

It is not only altering values on read but also on write when doing the `tdp plan reconfigure` command. Yes, it is possible to put the `filter_active` to True in the script of the`tdp plan reconfigure` command. However, the reason I put it in the query `get_hosted_entity_statuses` is that I thought it would always be illogical to restart or configure a component which is on a host that does not exist anymore so it was logical to condition the values of these two columns on whether the component actually exists or not.

PaulFarault:
Ok, just use the `filter_active` where it's needed then!

SteBaum:
> Ok, just use the `filter_active` where it's needed then!

Done

PaulFarault (approved):

SteBaum:
is_active column has been removed

PR #608, opened 2024-07-03
fix: deploy exception playbooks always on all hosts

SteBaum:

Fixes None

#### Additional comments

This is one approach to solve the issues of playbooks which always have to be deployed on all hosts even if only a specific one is set in the command. It is the case for the following playbooks in the tdp collection:
- tdp/playbooks/hbase_kerberos_install.yml
- tdp/playbooks/hdfs_kerberos_install.yml
- tdp/playbooks/yarn_kerberos_install.yml
- tdp/playbooks/hadoop_client_config.yml

#### Agreements

### DIFF
--- a/tdp/core/deployment/executor.py
+++ b/tdp/core/deployment/executor.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import Optional
 
 from tdp.core.models.enums import OperationStateEnum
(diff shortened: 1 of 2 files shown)


PaulFarault (commented):
Let's discuss about the way we could solve this issue with @rpignolet at next meeting.

PR #614, opened 2024-07-18
fix: only generate stales for deployed services

### LINKED ISSUE #613: Generate stales only for existing services
At the moment if we deploy only the basic tdp-collection but edit the vars for instance by mentioning in addition the tdp-extra collection, some components of the tdp-extra collection which do not exist on the cluster will have to be restarted or configured according to the `generate_stale_sch_logs` function which will always fail when launching a deployment with previously `tdp plan reconfigure`.

This is wrong and it should only be the other way around. if several collections are deployed we can chose to only reconfigure the services of one collection instead of all and for this reason the `--collection-path` option comes in handy if all collections have been defined in a path in an .env file. Therefore all potential restarted or configured services/components must exist before and have to be checked for it.

SteBaum:

Fixes #613

#### Additional comments

With this PR stales are only generated for services/components which previously exist in the database.

#### Agreements

### DIFF
--- a/tdp/cli/commands/status/generate_stales.py
+++ b/tdp/cli/commands/status/generate_stales.py
@@ -51,7 +51,7 @@ def generate_stales(
     check_services_cleanliness(cluster_variables)
 
     with Dao(db_engine) as dao:
-        stale_status_logs = dao.get_cluster_status().generate_stale_sch_logs(
+        stale_status_logs = dao.generate_stale_sch_logs(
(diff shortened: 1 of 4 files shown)


SteBaum:
With last force push moved `generate_stale_sch_logs` function to DAO.

SteBaum:
With last force push separated into two commits for easier understanding.

PaulFarault:
> With last force push moved `generate_stale_sch_logs` function to DAO.

Why?

PR #617, opened 2024-07-26
`browse` the last deployments

### LINKED ISSUE #616: `tdp browse` should show last X deployments instead of first X
When you have more deployments than the limits (by default 15), `tdp browse` only shows the first X deployments (X being the limit).

It should show the last X deployments. Last deployment should be on bottom of list.

SteBaum:

Fixes #616

#### Additional comments

Changed the order in query. Now showing the deployments in descending order.

#### Agreements

### DIFF
--- a/tdp/cli/commands/browse.py
+++ b/tdp/cli/commands/browse.py
@@ -89,7 +89,7 @@ def browse(
             return
 
         # Print all deployments
-        _print_deployments(dao.get_deployments(limit=limit, offset=offset))
+        _print_deployments(dao.get_last_deployments(limit=limit, offset=offset))
 
 
 def _print_deployments(deployments: Iterable[DeploymentModel]) -> None:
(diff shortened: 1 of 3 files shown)


PACordonnier:
Thanks @SteBaum. I just tried and it works. I reconsider what I said during the meeting and would rather have the most recent deployment at the end of the output, it makes more sense in a CLI. @rpignolet was right ! Can you modify it ?

Also, I think this is a fix rather than a refactor.

rpignolet:
We should talk about it with @PaulFarault how to fix that (find the good query to do the job)

SteBaum:
@PACordonnier see if it suits you now. The query returns the last deployments (15 by default) in ascending order.

PACordonnier:
Works perfect thanks

PACordonnier (approved):

PaulFarault:
> We should talk about it with @PaulFarault how to fix that (find the good query to do the job)

This is the correct query to modify. However, I'm unsure why a subquery is being introduced here.

PaulFarault (changes requested):
> We should talk about it with @PaulFarault how to fix that (find the good query to do the job)

This is the correct query to modify. However, I'm unsure why a subquery is being introduced here.

PaulFarault (approved):

rpignolet (approved):

PR #618, opened 2024-08-01
check if missing hosts during the deployment

SteBaum:

Fixes # None

#### Additional comments

When a deployment is launched check if hosts have been removed from the inventory.ini. The logger displays a warning if a host has been removed.

There are three commits:
- Added `get_collections_hosts` method for the `Collections` class.
- Added `check_collections_hosts` method for the `Collection` class. It compares hosts from the collections with the hosts in the `inventory.ini` file and logger sends a warning if a host is missing in the `inventory.ini` file.
- Integrated `check_collections_hosts` in the deployment iterator.

#### Agreements

### DIFF
--- a/tdp/core/collections.py
+++ b/tdp/core/collections.py
@@ -19,6 +19,7 @@ from collections.abc import Mapping, Sequence
 from tdp.core.collection import Collection
 from tdp.core.entities.hostable_entity_name import ServiceComponentName
 from tdp.core.entities.operation import Operations
+from tdp.core.inventory_reader import InventoryReader
 from tdp.core.operation import Operation
 from tdp.core.variables.schema.service_schema import ServiceSchema
 
@@ -242,3 +243,13 @@ class Collections(Mapping[str, Collection]):
             if operation.service_name == service_name
             and not operation.is_service_operation()
         }
+
+    def check_collections_hosts(self):
+        """Checks if hosts has not been removed from the inventory.ini file"""
+        inventory = InventoryReader().inventory
+        inventory.refresh_inventory()
+
+        inventory_list = [str(name) for name in inventory.get_hosts()]
+        for host in self.get_collections_hosts():
+            if host not in inventory_list:
+                return host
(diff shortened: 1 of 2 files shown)


tdp/cli/commands/deploy.py
```diff
@@ -82,7 +82,11 @@ def deploy(
             ),
             cluster_variables=cluster_variables,
             cluster_status=dao.get_cluster_status(),
-        ).run(planned_deployment, force_stale_update=force_stale_update)
+        ).run(
+            planned_deployment,
+            force_stale_update=force_stale_update,
+            ansible_inventory_file=str(run_directory) + "/inventory/inventory.ini",
```
rpignolet, line 88: You can not hard code the inventory file to read. This is why we have all the boilerplate inside `InventoryReader` to let Ansible parse config files and load all inventory files. The downside is inventory files are read when `ansible` module is imported and inventory is read once.

SteBaum: Yes you are right, I did all these because I did not see that there was a `refresh_inventory()` module for the InventoryManager. The code has been refactored and simplified.

SteBaum:
The host checking procedure slows down the deployment. It can specifically be seen when a `tdp deploy --dry` or `tdp -deploy --mock-deploy` is performed.

PaulFarault (changes requested):
I see several issues here.

1. `get_collections_hosts` should be a property initialized directly within the class, rather than a getter that computes it each time it's called. Additionally, its implementation can be made much more efficient by creating a set directly, avoiding the need for three loops.
2. `check_collections_hosts` is currently returning the first host removed from the collections compared to the inventory:

  - This behavior is unrelated to the operations running in the current deployment.
  - Since it always returns the same value, there’s no need to invoke it for every operation.

It could be useful to both store the hosts used in the collections and verify that the hosts for the current operation haven't been removed from `inventory.ini`. However, the current approach doesn't seem like the right solution for this.

SteBaum:
No more up to date.

PR #619, opened 2024-08-02
fix: seperated _update_cluster_fn from _process_operation_fn to displ…

SteBaum:
fix: seperated _update_cluster_fn from _process_operation_fn to display deployment operations

Fixes None

#### Additional comments

Before this PR the `tdp deploy` or `tdp deploy --mock-deploy` does not display the `-install` operations although they are executed and written in the database since they do not provide cluster_status_logs directly. Therefore in order to show these commands too, in the iterator the _process_operation_fn must be separated in two. one function which executes the operation and a second one which updates the status_logs.

#### Agreements

### DIFF
--- a/tdp/cli/commands/deploy.py
+++ b/tdp/cli/commands/deploy.py
@@ -98,11 +98,12 @@ def deploy(
         dao.session.commit()  # Update operation status to RUNNING
         for operation_rec, process_operation_fn in deployment_iterator:
             dao.session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
-            if process_operation_fn and (cluster_status_logs := process_operation_fn()):
+            if process_operation_fn:
                 click.echo(
                     f"Operation {operation_rec.operation} is {operation_rec.state} {'for hosts: ' + operation_rec.host if operation_rec.host is not None else ''}"
                 )
-                dao.session.add_all(cluster_status_logs)
+                if cluster_status_logs := process_operation_fn():
+                    dao.session.add_all(cluster_status_logs)
             dao.session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
 
         if deployment_iterator.deployment.state != DeploymentStateEnum.SUCCESS:


PaulFarault (changes requested):
If I understand correctly, you're saying that the CLI doesn't display any message in the console for the `*_install` operations because their `_process_operation_fn` aren't returning any logs to append to the `sch_status_log table`.

If so, extracting the log generation logic from the `_process_operation_fn` is a good refactor. However, the `DeploymentIterator` signature likely shouldn't change. Why return two functions instead of simply doing the following:

```diff
for operation_rec, process_operation_fn in deployment_iterator:
    dao.session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
-   if process_operation_fn and (cluster_status_logs := process_operation_fn()):
+   if process_operation_fn:
        click.echo(
            f"Operation {operation_rec.operation} is {operation_rec.state} {'for hosts: ' + operation_rec.host if operation_rec.host is not None else ''}"
        )
+       if cluster_status_logs := process_operation_fn():
-       dao.session.add_all(cluster_status_logs)
+           dao.session.add_all(cluster_status_logs)
    dao.session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
    

            dao.session.add_all(cluster_status_logs)
    dao.session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
```

SteBaum:
> If I understand correctly, you're saying that the CLI doesn't display any message in the console for the `*_install` operations because their `_process_operation_fn` aren't returning any logs to append to the `sch_status_log table`.
> 
> If so, extracting the log generation logic from the `_process_operation_fn` is a good refactor. However, the `DeploymentIterator` signature likely shouldn't change. Why return two functions instead of simply doing the following:
> 
> ```diff
> for operation_rec, process_operation_fn in deployment_iterator:
>     dao.session.commit()  # Update deployment and current operation status to RUNNING and next operations to PENDING
> -   if process_operation_fn and (cluster_status_logs := process_operation_fn()):
> +   if process_operation_fn:
>         click.echo(
>             f"Operation {operation_rec.operation} is {operation_rec.state} {'for hosts: ' + operation_rec.host if operation_rec.host is not None else ''}"
>         )
> +       if cluster_status_logs := process_operation_fn():
> -       dao.session.add_all(cluster_status_logs)
> +           dao.session.add_all(cluster_status_logs)
>     dao.session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
>     
> 
>             dao.session.add_all(cluster_status_logs)
>     dao.session.commit()  # Update operation status to SUCCESS, FAILURE or HELD
> ```

The seperation makes sens since executing the operations and creating the status logs are two different operations. However I have modified as you said to keep it as close as what was before and if the separation takes place it will be n another PR.

PaulFarault (approved):

rpignolet (approved):

PR #620, opened 2024-08-08
fix: get rid of red error marker in get_components_from_service

SteBaum:

Fixes None

#### Additional comments

- Get rid of red error marker in get_components_from_service in Collections class.
- Get rid of red error marking in Dag class.

#### Agreements

### DIFF
--- a/tdp/core/dag.py
+++ b/tdp/core/dag.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING, Optional, TypeVar
 import networkx as nx
 
 from tdp.core.constants import DEFAULT_SERVICE_PRIORITY, SERVICE_PRIORITY
+from tdp.core.entities.operation import Operations
 from tdp.core.operation import Operation
 
 if TYPE_CHECKING:
@@ -50,7 +51,7 @@ class Dag:
         self._graph = self._generate_graph(self.operations)
 
     @property
-    def operations(self) -> dict[str, Operation]:
+    def operations(self) -> Operations:
         """DAG operations dictionary."""
         return self._operations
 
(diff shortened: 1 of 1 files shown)


PaulFarault (changes requested):
Agree with the `Dag` fix.

For the `Collections.get_components_from_service`, we may just want to remove the `Operation.is_service_operation` method entirely and replace it with what you suggested (checking the `Operation.component_name` value).

SteBaum:
> Agree with the `Dag` fix.
> 
> For the `Collections.get_components_from_service`, we may just want to remove the `Operation.is_service_operation` method entirely and replace it with what you suggested (checking the `Operation.component_name` value).

Yes both are mutually exclusive so we do not need both. I removed `Operation.is_service_operation` with last force push.

PaulFarault (approved):

tdp/core/dag.py
```diff
@@ -274,28 +275,29 @@ def warning(collection_name: str, message: str) -> None:
...
-            if (
-                dependency.endswith("_start")
-                and dependency_service != operation.service_name
-            ):
-                c_warning(
-                    f"Operation '{operation_name}' is in service '{operation.service_name}', depends on "
-                    f"'{dependency}' which is a start action in service '{dependency_service}' and should "
-                    f"only depends on start action within its own service"
-                )
+        if operation.collection_name is not None:
```
rpignolet, line 278: `collection_name` is optional for an operation ?

SteBaum: At the moment yes. In the class `Operation`:

```
def __init__(
        self,
        name: str,
        collection_name: Optional[str] = None,
```

PaulFarault: Maybe you could invert the condition (using `continue`) to improve code and diff.

rpignolet: I understand you want to fix red marker but these marker remind us that the code need refactor. With this fix we will not remember to remove this `if` when `collection_name` will be mandatory.

PaulFarault:
I suggest we close this PR that only fixes the symptoms and not the main issues.

I opened #625, we'll need to discuss if we should make the `collection_name` mandatory for the `Operation`.

PR #621, opened 2024-09-09
feat: diplays the repo if GitError

SteBaum:

Fixes None

#### Additional comments

If an error occurs with the `is_clean` function of the `GitRepository` class, the concerned repository is printed in the trace.

#### Agreements


PaulFarault (commented):
Printing the source repository seems reasonable.

However, could you please provide an example of a case where this exception is raised?

SteBaum:
Yes sorry I should have copied the error I had. I had this error when I was executing TDP-lib in a container with a different user as on my host where tdp_vars were located and attached to the container through a volume. It could not access any git repository in this case. it had nothing to do with the code in tdp-lib and all git repositories were affected but in the case that only one would be affected we would not know which one it is, so this eror message tells us precisely which one is concerned.

PaulFarault:
Then I'm not sure that this PR is that useful. We never encountered this type of error and even if it occurs, your PR it only covering the `is_clean` method.

PR #628, opened 2024-10-22
Improve dag parsing logic

PaulFarault:

Fixes None

#### Additional comments

- Isolate the DAG parsing logic from the collection module.
- `Collection.dag_nodes` now returns a generator.

#### Agreements


tdp/core/collection/collection.py
```diff
@@ -125,9 +117,9 @@ def schema_directory(self) -> Path:
         return self._path / SCHEMA_VARS_DIRECTORY_NAME
 
     @property
-    def dag_nodes(self) -> list[TDPLibDagNodeModel]:
-        """List of DAG files in the YAML format."""
-        return self._dag_nodes  # TODO: should return a generator
+    def dag_nodes(self) -> Generator[TDPLibDagNodeModel, None, None]:
+        """DAG nodes."""
```
PaulFarault, line 121: Should change it to "Read DAG nodes." to highlight that the Generator is created when calling the method.

PaulFarault: Done.

tdp/core/collection/collection.py
```diff
@@ -125,9 +117,9 @@ def schema_directory(self) -> Path:
         return self._path / SCHEMA_VARS_DIRECTORY_NAME
 
     @property
-    def dag_nodes(self) -> list[TDPLibDagNodeModel]:
-        """List of DAG files in the YAML format."""
-        return self._dag_nodes  # TODO: should return a generator
+    def dag_nodes(self) -> Generator[TDPLibDagNodeModel, None, None]:
+        """Read DAG nodes."""
+        return parse_dag_directory(self.dag_directory)
```
rpignolet, line 122: You change the logic here, with the new logic, every call to `dag_nodes` read the filesystem which is not what we want. The filesystem should be read once when we populate `Collections` and keep all information read in memory.

PaulFarault: I did this on purpose. `Collections` is indeed calling `Collection.dag_nodes` only once to initialize itself.

IMO, the `Collection` class should be used to get the **current** collection content. I would like to accomplish the same for the other properties. I planed to do that in another PR but it may makes more sense to do everything at once.

PaulFarault: That way, we would be able to use the `Collection` class to update the `Collections` object when needed without having to recreate each `Collection` objects.

rpignolet: haha now you want to have a `Collection` which have content updated for each function call whereas you are the one who wanted to have immutable classes so I don't understand.

PaulFarault: Yeah, I kinda changed my mind for `Collection` 💁🏻‍♂️

PaulFarault: Fixed it in the new version : I removed the `@property` (except for directories, name and path) and replace them with methods `read_dag_nodes`, `read_playbooks` and `read_schemas`.

PaulFarault:
This new version transforms the `Collection` class to `CollectionReader`.

The `CollectionReader` class doesn't store data anymore, only a few metadata remain (collection `name`, `path` and paths to directories). Instead, I introduced 3 methods that directly read from the collection on call : `read_dag_nodes`, `read_playbooks` and `read_schemas`. Now, the use of `CollectionReader is strictly limited to `Collections` that uses it to initialize itself.

`Collections` and `CollectionReader` has been move to a `tdp.core.collections` package that only exports the `Collections` class and a few `Error` that can be raised. Methods used by the readers methods from the `CollectionReader` class has been extracted into their own files and unit tests have been refactored accordingly.

rpignolet (approved):

PR #652, opened 2025-07-07
Refactor CLI params definitions

PaulFarault:

Fixes None

#### Additional comments

Refactor the CLI's parameters definitions:

- Remove the barrel file (`__init__.py` module that was re-exporting params).
- Gather parameters definitions into a single `tdp/cli/params.py` module
- Import `core` logic in option definition. For example:

    ```py
    def collections_option(func: FC) -> FC:
        from tdp.core.collections import Collections

        return click.option(
            # ...
            callback=lambda ctx, param, value: Collections.from_collection_paths(value)
        )(func)
    ```

    Instead of:

    ```py
    from tdp.core.collections import Collections
    
    def _collections_from_paths(
        ctx: click.Context, param: click.Parameter, value: list[Path]
    ) -> Collections:
        if not value:
            raise click.BadParameter("cannot be empty", ctx=ctx, param=param)
        return Collections.from_collection_paths(value)
    
    
    def collections_option(func: FC) -> FC:
        return click.option(
            # ...
            callback=_collections_from_paths
        )(func)
    ```

- Extract some logic used in params callback to define in in store (more specifically, `tdp/core/db.py` module used by the `--database-dsn` option).

`component_argument_option` and `service_argument_option` arguments callbacks also need to be simplified but they will probably need to be rewritten to make use of the new `Collections.entities` which will simplify their definition.

#### Agreements

### DIFF
--- a/scripts/playbooks.py
+++ b/scripts/playbooks.py
@@ -20,7 +20,7 @@ from pathlib import Path
 import click
 import networkx as nx
 
-from tdp.cli.params.collections_option import collections_option
+from tdp.cli.params import collections_option
 from tdp.core.constants import DEFAULT_SERVICE_PRIORITY, SERVICE_PRIORITY
 from tdp.core.dag import Dag
 from tdp.core.entities.operation import OperationName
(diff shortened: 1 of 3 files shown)


PaulFarault:
Force pushed after rebase on `master`.

PR #657, opened 2025-07-10
Rename `--overrides` option to `--conf`

PaulFarault:

Fixes None

#### Additional comments

Rename the `--overrides` option to `--conf` to clarify the fact that user configuration should be passed through this option.

#### Agreements

### DIFF
--- a/tdp/cli/commands/init.py
+++ b/tdp/cli/commands/init.py
@@ -14,7 +14,7 @@ from tdp.cli.params import (
     validate_option,
     vars_option,
 )
-from tdp.cli.params.overrides_option import overrides_option
+from tdp.cli.params.conf_option import conf_option
 
 if TYPE_CHECKING:
     from sqlalchemy import Engine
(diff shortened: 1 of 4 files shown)


tdp/cli/commands/init.py
```diff
@@ -23,13 +23,13 @@
...
 @click.command()
-@overrides_option
+@conf_option
 @collections_option
 @database_dsn_option
 @validate_option
 @vars_option(exists=False)
 def init(
-    overrides: tuple[Path],
+    user_vars: tuple[Path],
```
rpignolet, line 32: Conf option map to `user_vars`, why ??? It is confusing.

PaulFarault: `--conf` option naming is only scoped to the `cli` package. Once in the `core` package, I used `user_vars` everywhere which goes very well with `vars`.

IMO it would be confusing to call it `conf` in the `core` package. However, I can call use `conf` for Click parameter injection if this is what you find confusing.

rpignolet: I think it should be call `conf` everywhere to avoid confusion like before was called `overrides` and the word `override` is in the core.

PaulFarault: `conf` is probably the most confusing naming we can use to designate user variables in `core`. Why not using something specific to designate what it really is?

PaulFarault:
Revert the renaming inside of `core`.

PR #678, opened 2025-10-16
feat(scripts): the playbooks script now makes start.yml, restart.yml …

SteBaum:
… chosen host

Fixes None

#### Additional comments

The PRs [tdp-collection#989](https://github.com/TOSIT-IO/tdp-collection/pull/989) an [tdp-collection-extras#236(https://github.com/TOSIT-IO/tdp-collection-extras/pull/236) are the output of this script.

#### Agreements

### DIFF
--- a/scripts/playbooks.py
+++ b/scripts/playbooks.py
@@ -154,6 +154,53 @@ def playbooks(
             else:
                 all_fd.write(f"# {operation.name}\n")
 
+    with Path(meta_dir, "start.yml").open("w") as start_fd:
+        write_copyright_licence_headers(start_fd)
+        start_fd.write("---\n")
+        for operation in dag.get_all_operations(restart=False, stop=False):
+            if for_collection and not is_in_collection(operation):
+                continue
+            if is_noop(operation):
+                start_fd.write(f"## {operation.name}\n")
+            elif "start" == operation.name.action:
+                start_fd.write(
+                    f"- ansible.builtin.import_playbook: {playbooks_prefix}{operation.name}.yml\n"
+                )
+            else:
+                start_fd.write(
+                    f"# - ansible.builtin.import_playbook: {playbooks_prefix}{operation.name}.yml\n"
+                )
+
+    with Path(meta_dir, "restart.yml").open("w") as restart_fd:
+        write_copyright_licence_headers(restart_fd)
+        restart_fd.write("---\n")
+        for operation in dag.get_all_operations(restart=True, stop=False):
+            if for_collection and not is_in_collection(operation):
+                continue
+            if is_noop(operation):
+                restart_fd.write(f"## {operation.name}\n")
+            elif "restart" == operation.name.action:
+                restart_fd.write(
+                    f"- ansible.builtin.import_playbook: {playbooks_prefix}{operation.name}.yml\n"
+                )
+            else:
+                restart_fd.write(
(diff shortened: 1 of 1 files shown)


scripts/playbooks.py
```diff
@@ -154,6 +154,45 @@ def is_in_collection(operation: DagOperation) -> bool:
...
+        start_fd.write("---\n")
+        for operation in dag.get_all_operations(restart=False, stop=False):
+            if for_collection and not is_in_collection(operation):
+                continue
+            if not is_noop(operation) and "start" in operation.name.action:
+                start_fd.write(
+                    f"- ansible.builtin.import_playbook: {playbooks_prefix}{operation.name}.yml\n"
+                )
+            else:
+                start_fd.write(f"# {operation.name}\n")
```
rpignolet: Can you prefix 2 `#` for noop operation and 1 `#` for "normal" operation. This way we can see what operation is noop.

scripts/playbooks.py
```diff
@@ -154,6 +154,45 @@ def is_in_collection(operation: DagOperation) -> bool:
...
+                restart_fd.write(
+                    f"- ansible.builtin.import_playbook: {playbooks_prefix}{operation.name}.yml\n"
+                )
+            else:
+                restart_fd.write(f"# {operation.name}\n")
+
+    with Path(meta_dir, "stop.yml").open("w") as stop_fd:
+        write_copyright_licence_headers(stop_fd)
+        stop_fd.write("---\n")
+        for operation in dag.get_all_operations(restart=False, stop=True):
```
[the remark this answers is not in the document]
SteBaum: You are right, it has to be reversed for the stop order between the components. However, in this case the order is also reversed within the components and therefore the playbook order ends with the `install` playbooks which makes no sense. I think that all commented playbooks should be removed in this case since it makes no sense.

PaulFarault (changes requested):
Should this be optimized?

We're iterating over the dag operations many times. It would may be better to only iterate once and then perform the writing in the files.

scripts/playbooks.py
```diff
@@ -154,6 +154,49 @@ def is_in_collection(operation: DagOperation) -> bool:
...
                 all_fd.write(f"# {operation.name}\n")
 
+    with Path(meta_dir, "start.yml").open("w") as start_fd:
+        write_copyright_licence_headers(start_fd)
+        start_fd.write("---\n")
+        for operation in dag.get_all_operations(restart=False, stop=False):
+            if for_collection and not is_in_collection(operation):
+                continue
+            if is_noop(operation):
+                start_fd.write(f"# {operation.name}\n")
```
PaulFarault: Aren't noop supposed to be the one prefixed with `##`?

PaulFarault (changes requested):

SteBaum:
> > Yes but should the noop operations in other playbooks also be prefixed with ## then if we want to be consistent through out all metas ?
> 
> There is no need for 2 `#` for the other, if a user use a massive insert à first position of all lines, it will add the second `#` for noop.
> 
> On the other hand, for playbooks that are not noop, the entire line must work, not just the name of the operation, the line must be like this
> 
> ```
> # - ansible.builtin.import_playbook: ../zookeeper_server_install.yml
> ```

Done

PaulFarault (approved):

PR #682, opened 2025-11-13
Feat/all host operations

SteBaum:
#### Which issue(s) this PR fixes

Fixes None

#### Additional comments

Puts an optional parameter which erases the host limit for a specified operation when a host limit is set.

#### Agreements

- [x] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0).
- [x] I hereby agree to grant [TOSIT](https://www.tosit.io/) a copyright license to use my contributions.

PR #683, opened 2025-11-17
Restrict `ansible-playbook`'s `--limit` option on specific operations

PaulFarault:

Alternative to #679 and #682

#### Additional comments

- **Introduce playbook metadata**.  
  Metadata for a playbook can now be defined in any play, using the `vars.tdp_lib` dictionary. When multiple plays provide metadata, the lib will merges them.

  **Example**:

  ```yaml
  - name: task 1
    vars:
      tdp_lib:
        meta1: false

  - name: task 2

  - name: task 3
    vars:
      tdp_lib:
        meta2: foo
  ```

  **Resulting metadata**:

  ```
  meta1: false
  meta2: foo
  ```

  Overriding rules will be defined on a per-metadata basis. 

  For now, only the `can_limit` metadata is supported. It defaults to `true`.  
  Setting `can_limit: false` indicates that the play must run on _all_ hosts (i.e., cannot be restricted using `ansible-playbook --limit`). This is required for certain global operations (e.g. Kerberos initialization).  
  Specifying `can_limit: false` on _any_ play applies to the entire playbook. If another play within the same playbook explicitly sets `can_limit: true`, a warning will be logged.
- **Refactored playbook reading and validation**.  
  Playbook parsing and validation have been moved from `InventoryReader` to a new `playbook_validate` module. This make validation reusable (notably for the metadata extraction). The validation logic now relies on Pydantic rather than manual checks.
- **Clearer deployment error messages**.  
  When running an operation with `--limit`, error messages are now more explicit. They distinguish between:

  1. operations that can't be executed (noop);
  2. operations not applicable to the targeted hosts;
  3. operations that cannot be limited.

  This will make it easier to introduce differentiated handling for these cases later.

#### Agreements

### DIFF
--- a/tdp/core/collections/collection_reader.py
+++ b/tdp/core/collections/collection_reader.py
@@ -7,11 +7,12 @@ import json
 import logging
 from collections.abc import Generator
 from pathlib import Path
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 import yaml
 from pydantic import BaseModel, ConfigDict, ValidationError
 
+from tdp.core.collections.playbook_validate import validate_playbook
 from tdp.core.constants import (
     DAG_DIRECTORY_NAME,
     DEFAULT_VARS_DIRECTORY_NAME,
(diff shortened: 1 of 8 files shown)


PaulFarault:
The PR has been completed.

Note that the current implementation only prevent from using the `--limit` option when a `can_limit` meta is defined for a playbook. It does **NOT** prevent the CLI to generate deployment plan with a limit on `can_limit: false`. This need to be fixed in another PR.

